PR #536, opened 2013-04-14
Since there was a bug fix that does not specify a mailbox at delete_all

taketin:
I was using the delete_all but errors are out "(Error in IMAP command received by server.) Net :: IMAP :: NoResponseError", was added because there was no process to select the mailbox.


jeremy:
Fixed by 4d1319696f48d9a2a010f81cb18ef741b1d89dae

PR #565, opened 2013-06-18
Message should be able to parse multipart messages that rely on the Boundary for delimiting header and body

tylergannon:
Hi!  I hope this change will come as agreeable to you... I am working on a production app that needs to be able to receive and parse emails from a client that is putting white space in the header of multipart messages.

There are new tests added for this case, and everything's green.  Please let me know if there's anything else you'd like me to do in order to get this pushed out... I would be really pleased if we could merge this upstream and make a release!

---

Multipart messages some times arrive with whitespace in the header.  In
this case, the Boundary specified in the Content-Type header determines
the end of the header section and the beginning of the body.

This commit handles that case.


jeremy:
Fixed by #539

PR #568, opened 2013-06-21
Parse UTF8 and ISO-8859-1 text in email addresses.

peterkovacs:
I modified common.rl to include UTF8 sequences as defined in http://datatracker.ietf.org/doc/rfc6532/?include_text=1.

I also extended `ALPHA` and `atext` to include 0xa0..0xff -- thus enabling support for iso-8850-1 encoded text.

This is a result of seeing addresses like "françois@example.com" and "fran\xe7ois@example.com" in the wild.

I understand that by adding the high bits from iso-8859, this is no longer RFC compliant, but for our application (at SaneBox) its better to be much more accepting of data from the wild than it is to reject it.


jeremy:
UTF-8 header parsing is supported in #1103

PR #579, opened 2013-07-08
Do not attempt to collapse broken encodings in fields

bpot:
Without this strings with broken encodings will raise an exception in `Encodings.collapse_adjacent_encodings:` `NoMethodError: undefined method`+' for nil:NilClass`


jeremy:
Fixed by #789

PR #586, opened 2013-07-23
Access message parts via IMAP's part specifiers

benhamill:
In RFC 3501, it talks about how you can request just a specific part of a message using numbers separated by dots. So '1.2.1' would get the first part of the second part of the first part. For messages that aren't multipart '1' means the message its self. I thought it would be handy to have a way to fetch message parts by the specifier, so I wrote a test and an implementation.

Is this something people are interested in? Does the code style and testing methodology look right in the larger context of the project? Should I be raising `ArgumentError` instead of returning `nil` in the places where I currently do so?

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -1955,6 +1955,24 @@ module Mail
       has_content_type? ? !!(main_type =~ /^text$/i) : false
     end
 
+    # Returns the part of the body as described by the IMAP part specifier in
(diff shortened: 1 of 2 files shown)


jeremy:
This is cool! But pretty obscure for most users. Thanks @benhamill 😊

benhamill:
Haha. Fair enough. Four years ago is pretty... ago.

PR #587, opened 2013-07-24
return header fields if Message#headers called without argument

defeated:
Recently I had to parse specific headers from an incoming email and was "surprised" at the behavior of the `Message#headers` method.

This small change aligns the api of `headers` so that it behaves similarly to the `header` method: if no argument is passed to headers, return the header fields.


jeremy:
Moving toward decreasing the API footprint rather than spreading it further. Would rather deprecate and remove 😊

PR #591, opened 2013-07-29
Fix forgot to replace at 702b306886bae732f2c6c15860320e20143b0059

mrkn:
This fixes the commit 702b306886bae732f2c6c15860320e20143b0059 to forget to replace a `raw_source.to_s` with `raw_string`.


kennethkalmer:
@jeremy @mikel @mrkn I had a quick look and it seems this is still perfectly valid!

Issue #564, opened 2013-06-13
Date parser fails with invalid time: 30:00


gauteh:
The date parse fails if the Date field in the raw source is of the kind:

```
Date:   Wed, 7 Nov 2007 30:00:45 +0700
```

With the following error:

```
/home/gaute/dev/ruby/mail/lib/mail/parsers/ragel/ruby/machines/date_time_machine.rb:576:in `parse': undefined method `length' for nil:NilClass (NoMethodError)
    from /home/gaute/dev/ruby/mail/lib/mail/parsers/ragel/ruby.rb:24:in `parse'
    from /home/gaute/dev/ruby/mail/lib/mail/parsers/ragel.rb:9:in `parse'
    from /home/gaute/dev/ruby/mail/lib/mail/parsers/date_time_parser.rb:10:in `parse'
    from /home/gaute/dev/ruby/mail/lib/mail/elements/date_time_element.rb:8:in `initialize'
    from /home/gaute/dev/ruby/mail/lib/mail/fields/common/common_date.rb:34:in `new'
    from /home/gaute/dev/ruby/mail/lib/mail/fields/common/common_date.rb:34:in `element'
    from /home/gaute/dev/ruby/mail/lib/mail/fields/common/common_date.rb:6:in `date_time'
    from /home/gaute/dev/ruby/mail/lib/mail/fields/common/common_date.rb:10:in `default'
    from /home/gaute/dev/ruby/mail/lib/mail/field.rb:183:in `method_missing'
    from /home/gaute/dev/ruby/mail/lib/mail/message.rb:1187:in `default'
    from /home/gaute/dev/ruby/mail/lib/mail/message.rb:592:in `date'
    from testmail.rb:10:in `<main>'
```

Loading the mail using:

```
r = File.read 'mail'
m = Mail.read_from_source r
m.date
```

jeremy:
#594

Issue #216, opened 2011-03-16
invalid date throws error


dgm:
When processing a message that has an invalid date, I had a rails page crash.  Does Mail do this for all parsing errors, or can it catch the errors and return the best parse it can?  The nature of my project (Maia Mailguard) is such that I am sure to have broken messages that I'm trying to look at.  If you want edge cases to test against, I'm sure I can find them :)

invalid date

```
/Users/dgm/.rvm/rubies/ruby-1.9.2-p0/lib/ruby/1.9.1/date.rb:1688:in `new_by_frags'
/Users/dgm/.rvm/rubies/ruby-1.9.2-p0/lib/ruby/1.9.1/date.rb:1733:in`parse'
mail (2.2.9) lib/mail/fields/date_field.rb:41:in `initialize'
mail (2.2.9) lib/mail/field.rb:176:in`new'
mail (2.2.9) lib/mail/field.rb:176:in `new_field'
mail (2.2.9) lib/mail/field.rb:145:in`create_field'
mail (2.2.9) lib/mail/field.rb:74:in `initialize'
mail (2.2.9) lib/mail/header.rb:78:in`new'
mail (2.2.9) lib/mail/header.rb:78:in `block in fields='
mail (2.2.9) lib/mail/header.rb:76:in`each'
mail (2.2.9) lib/mail/header.rb:76:in `fields='
mail (2.2.9) lib/mail/header.rb:252:in`split_header'
mail (2.2.9) lib/mail/header.rb:40:in `initialize'
mail (2.2.9) lib/mail/message.rb:364:in`new'
mail (2.2.9) lib/mail/message.rb:364:in `header='
mail (2.2.9) lib/mail/message.rb:1774:in`parse_message'
mail (2.2.9) lib/mail/message.rb:1903:in `init_with_string'
mail (2.2.9) lib/mail/message.rb:124:in`initialize'
mail (2.2.9) lib/mail/mail.rb:50:in `new'
mail (2.2.9) lib/mail/mail.rb:50:in`new'
mail (2.2.9) lib/mail/mail.rb:187:in `read_from_string'
```

jeremy:
#593

PR #593, opened 2013-08-05
Handle invalid date in Received header

okkez:
Raise ArgumentError If invalid date in Received header.
On the other hand, nothing raised if invalid date in Date header.
So I think mail library should not raise ArgumentError if invalid date in Received header.

This patch can improve robustness.
I created this patch for 2-5-stable branch but it may be applied for master.
(I have not checked for master yet.)

Thanks!

Related issues: #564, #216


jeremy:
Superseded by #594

PR #595, opened 2013-08-08
Iso 2022 jp - implement charset codec

yujiyokoo:
With help from @kuroda and @mikel, I have implemented the CharsetCodec which @jeremy suggested in the original PR (https://github.com/mikel/mail/pull/534).

I have simply moved things around - it should work exactly the same as the original iso-2022-jp branch AFAIK.

### DIFF
--- a/lib/mail/body.rb
+++ b/lib/mail/body.rb
@@ -27,9 +27,7 @@ module Mail
   class Body
 
     def initialize(string = '')
-      if string.respond_to?(:encoding) && string.encoding.to_s.downcase == 'iso-2022-jp'
-        string.force_encoding('US-ASCII')
-      end
+      CharsetCodec.force_regexp_compatibility_if_needed(string)
       @boundary = nil
       @preamble = nil
       @epilogue = nil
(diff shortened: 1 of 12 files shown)


thorsteneckel:
Hey @yujiyokoo, @kuroda and @mikel 👋 One of our users over at https://github.com/zammad/zammad stumbled across this issue [earlier this year](https://github.com/zammad/zammad/issues/2498) and now one of our paying customers is affected as well. Therefore this topic gets a certain priority and we were wondering what we can contribute to enable this great gem to support Iso 2022 jp mails. Could you please give us a short update on the current state of this PR - which seems to be staled for nearly 6 years by now. We'd like to pick things up and get this merged into the latest version. Looking forward do hear from you. Bests, Thorsten

kuroda:
@thorsteneckel 

The `mail-iso-2022-jp` is a hacked version of another gem that was maintained on https://github.com/ma2shita/mail_ja.

I managed to make this gem work, but I did not have a solid understanding of the internal structure.

So I couldn't answer the question from @jeremy.

I think it is very difficult to save this six year old PR.

yujiyokoo:
Yeah I think @kuroda is right. It's very hard to save such an old PR.

Back then, I tried working on this for a while, and was intending to do more work on it but unfortunately other things got prioritised over it and I didn't know a lot about the internals of this either :(

PR #602, opened 2013-08-13
Fix NoMethodError in Part#inline? when Content-Dispostion header can't be parsed.

a1exsh:
Invalid byte sequence in Content-Disposition field make it parse as UnstructuredField.  Later, then someone calls Part.inline? (like Message#ready_to_send! does via Part#add_required_fiels) it can fail with NoMethodError on call to `disposition_type.'

### DIFF
--- a/lib/mail/part.rb
+++ b/lib/mail/part.rb
@@ -34,7 +34,7 @@ module Mail
     end
     
     def inline?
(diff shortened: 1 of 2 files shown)


jeremy:
Fixed by #849

PR #605, opened 2013-08-17
instantiate an empty Mail::Address object and call name

peterkovacs:
Fixes an exception raised by `Mail::Address.new( nil ).name`


knu:
This should really be merged.  No exception in the constructor, a NameError when `name` is called.

jeremy:
Merged @ 8b6c98e46cdb8e164e985bebb89b439b599905ff and backported to 2-6-stable @ 97e0c7a5587ace957cab20f84ac2ac0fe31ff00d

Issue #408, opened 2012-06-11
Corrupted binary parts


ebeigarts:
Currently the `\n` char gets replaced with `\r\n` in binary parts:

My current workaround is:

``` ruby
module Mail
  class Message
    # Avoids calling {String#to_crlf} on binary parts.
    def raw_source=(value)
      if value.include?("Content-Transfer-Encoding: binary")
        @raw_source = value
      else
        @raw_source = value.to_crlf
      end
    end
  end
end
```

You can reproduce this with a PDF:

```
--uuid:f2520735-6098-4586-93f3-019733fc1404+id=4
Content-ID: <http://tempuri.org/1/634746078367623835>
Content-Transfer-Encoding: binary
Content-Type: application/octet-stream

%PDF-1.3
....
```

jeremy:
Fixed by #1113

PR #609, opened 2013-08-21
Fixing #408 by not standarizing whitespace

malmckay:
There may be a better fix, that can standardize line breaks that are outside of the binary blob, but I feel this is a good step forward

Parsing binary encoded attachments still fails for most jpgs in 1.8.7 (1.9+ is fine). Without this patch most binary encoded attachments fail in all versions, so, again, I feel this is good step forward.

PS This is a repeat pull request, my previous one was from HEAD, which included other commits


jeremy:
Folded in to #1113 😊

Issue #596, opened 2013-08-11
Possibly false “Non US-ASCII detected” warning


sorbits:
Running the program below results in this warning:

```
Non US-ASCII detected and no charset defined.
Defaulting to UTF-8, set your own if this is incorrect.
```

I can add `charset=utf-8` to the root `Content-Type` to avoid the warning, but I don’t think this should be required, as the root content consists of mixed parts that each can have its own charset.

```
require 'mail'

mail = Mail.read_from_string(DATA.read)
puts mail.to_s

__END__
From: me@example.com
To: you@example.com
MIME-Version: 1.0
Content-Type: multipart/mixed; boundary="=_UniqueBoundary_F3518054-C272-4B32-954D-A47A78F5531A_="

--=_UniqueBoundary_F3518054-C272-4B32-954D-A47A78F5531A_=
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: 8bit

Using “fancy” quotation.

--=_UniqueBoundary_F3518054-C272-4B32-954D-A47A78F5531A_=
Content-Disposition: attachment; filename=Binary.o
Content-Type: application/octet-stream
Content-Transfer-Encoding: base64

aGVsbG8gd29ybGQK
--=_UniqueBoundary_F3518054-C272-4B32-954D-A47A78F5531A_=--
```

jeremy:
Works without warning now. Thanks for the report, @sorbits!

PR #642, opened 2013-12-09
Don't try to add charset parameter for non-text types.

bk2204:
Non-text/* types, such as multipart/\*, image/\*, and the like often contain non-text content or content containing data of multiple charsets.  Almost all those non-text/* types that can use a character set, such as application/xhtml+xml, can signal this data in-band without the need for an explicit header.

Additionally, the warning about adding a charset in this case is nonsensical for types such as multipart/mixed.  Omit the autogeneration of the charset parameter for non-text/* types.

This has been tested against Ruby 1.8, 1.9.3, and 2.0. In each case, the two new tests fail before the fix and all tests pass afterwards.

This also addresses issue #596.

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -2036,7 +2036,7 @@ module Mail
       add_required_message_fields
       add_multipart_mixed_header    if body.multipart?
       add_content_type              unless has_content_type?
-      add_charset                   unless has_charset?
+      add_charset                   if text? && !has_charset?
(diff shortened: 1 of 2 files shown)


afn:
Can this be merged?  The bug still exists in master...

afn:
@mikel Any plans to include this in the next release? Thanks!

jeremy:
:+1:

afn:
Thanks! :smile: And not to push my luck, but any chance of getting this released to RubyGems in the near future?

PR #643, opened 2013-12-09
Fix missing .should calls in tests.

bk2204:
Several of the message tests were missing their .should invocations, resulting in those tests not testing what they should (no pun intended).  Modify these tests to use the proper syntax, fix the regular expressions to match appropriately, and mark the two failing tests as pending.

I adjusted the regular expressions after testing sample fixups for each of the two tests I marked pending. The charset=US-ASCII test fails because the output actually has charset=UTF-8, and the attachment test fails because a charset _is_ generated for attachments. I left them pending because it wasn't clear to me what the correct behavior should be.

### DIFF
--- a/spec/mail/message_spec.rb
+++ b/spec/mail/message_spec.rb
@@ -1117,18 +1117,20 @@ describe Mail::Message do
         end
 
         it "should set the content type to text/plain; charset=us-ascii" do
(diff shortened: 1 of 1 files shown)


jeremy:
Fixed by #976

PR #649, opened 2013-12-19
Collapse adjacent =??B? words losslessly

Senjai:
Cherry Picked: b832faf6d05f85422dd921d3b97ed842436f22a8

This issue broke our code base when we upgraded to rails 3.2.16. Currently we're using a local fork, but I thought that this should still go into 2-5-stable as a patch.

Richard

### DIFF
--- a/lib/mail/encodings.rb
+++ b/lib/mail/encodings.rb
@@ -292,7 +292,6 @@ module Mail
 
         if encoding == previous_encoding
           line = results.pop + line
-          line.gsub!(/\?\=\=\?.+?\?[QqBb]\?/m, '')
         end
 
         previous_encoding = encoding
--- a/spec/mail/encodings_spec.rb
+++ b/spec/mail/encodings_spec.rb
@@ -154,6 +154,12 @@ describe Mail::Encodings do
       Mail::Encodings.value_decode(string).should == string
     end
 
+    it "should collapse adjacent words" do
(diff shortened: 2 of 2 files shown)


corlissc:
+1 for this merge request.  Currently, the code in the library has a major data integrity issue, as the encoded words being concatenated could drop the additional RFC-1342 encoded-words.  Any idea on when this will be merged and backported?

bf4:
@corlissc which version are you on?

cc @jeremy @mikel @grosser @jeremyevans

jeremyevans:
This appears to be a low risk and high reward backport, so I'd be in favor of merging it into `2-5-stable`.

corlissc:
@jeremyevans, currently we are using 2.5.4 (don't ask, it's not my code base).  Thanks for your prompt response, it is greatly appreciated.  Let me know if I can be of any assistance.

corlissc:
@bf4 @jeremyevans I'm not sure if the Encodings.collapse_adjacent_encodings method is really necessary at all.  Unless it is necessary to remove any linear white-space (space or horizontal tab) or newline characters between adjacent RFC-1342 encoded-words for your parser to properly parse the header line.  If this is the case then I would suggest simplifying the method to something like:
```
    def Encodings.collapse_adjacent_encodings(str)
      Array(str.split(/(\?=)\s*(=\?)/).join)
    end
```

For obvious reasons Base64 can't be collapsed without first decoding and re-encoding (which defeats the purpose of performance gains by collapsing in the first place, if that was the original intent.)  Quoted-Printable could be collapsed, but, is there truly any gain in performance based on the amount of additional processing (loops, logical comparisons, regex comparisons, etc...) that is necessary to collapse it in the first place?

Just a few thoughts.  Please let me know if I can be of any assistance.

Senjai:
Closing this as PR cleanup.

PR #650, opened 2014-01-07
Handle unstructured field with UTF-7 encoded-word

johngrimes:
Attempting to parse an email with the following header:

```
References: =?utf-7?B?PDEyMzRAdGVzdC5saW5kc2Fhci5uZXQ+?=
```

Will cause an error: `Encoding::ConverterNotFoundError: code converter not found (UTF-7 to UTF-8)`.

This can be fixed by checking for UTF-7 within the `b_value_decode` method and using Net::IMAP to decode the string first.


tdg5:
I'm not sure when it was fixed, but this appears to work on v2.6.3. It still gives a warning, but otherwise seems like this PR could be closed?
```ruby
::Mail::Encodings.value_decode("=?utf-7?B?PDEyMzRAdGVzdC5saW5kc2Fhci5uZXQ+?=")
# Encoding conversion failed code converter not found (UTF-7 to UTF-8)
"<1234@test.lindsaar.net>"
```

PR #675, opened 2014-02-22
Sort attachments to end

### LINKED ISSUE #655: Text attachments appear as message body in some mail clients
The default part sort order is [ "text/plain", "text/enriched", "text/html" ].  If a multipart email contains text and html parts as well as a text attachment, the text attachment will often appear before the text body in the message.

In some mail clients this results in the attachment being treated as the message body.

I'm not sure if this is something that *should* be fixed, but if it is, a simple solution would be to look for content-disposition=attachment when sorting parts.

I can send a pull request if this seems like a legitimate bug.

npickens:
Fixes #655.


npickens:
@jeremy Updated to resolve the merge conflict.

jeremy:
Squashed 89c78b77278366740b3a7e8d8047e0c0926effa2

aryeh-looker:
💪

PR #683, opened 2014-03-04
fix SMTP dot-stuffing

yyyc514:
Mail does not do proper dot-stuffing and some MTAs will completely fail when trying to send email with un-stuffed periods

### DIFF
--- a/lib/mail/network/delivery_methods/smtp.rb
+++ b/lib/mail/network/delivery_methods/smtp.rb
@@ -96,6 +96,7 @@ module Mail
     # The from and to attributes are optional. If not set, they are retrieve from the Message.
     def deliver!(mail)
       smtp_from, smtp_to, message = check_delivery_params(mail)
+      message = dot_stuff(message)
 
       smtp = Net::SMTP.new(settings[:address], settings[:port])
       if settings[:tls] || settings[:ssl]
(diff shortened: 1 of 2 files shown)


kennethkalmer:
Nice work, I saw this fix got released with 2.2.0, which means that apart from 1.9.3 not being supported from February 2015, there are still a few versions of Ruby to support for a while. I don't know how quickly folks will be upgrading to 2.2.0.

I think the fix @yyyc514 got applied to Ruby itself is the best place. Maybe this issue should just be closed off, and if someone finds it while searching for it they'll know they have two options:

* Upgrade to 2.2.0 or later
* Add a newline to the end of their message body

If this gets merged, I'm worried we'll have to add an expiry date to it...

@mikel wdyt?

yyyc514:
It went in before 2.2. It's in 2.1.5 for example. I think 2.1.3 is when it first popped up.

jeremy:
The Ruby fix was backported to 2.1 and 2.0 stable as well, with a 1.9.3 backport pending - hopefully hits before 1.9.3 EOL. An upstream fix is appropriate, but it certainly wouldn't hurt to provide one ourselves either.

yyyc514:
Thanks.

PR #689, opened 2014-03-12
Fix Mail::Exim, forgotten in 4875bc2b

jethrogb:
The parameters of Mail::Sendmail.call changed in 4875bc2b, but Mail::Exim, which inherits from Mail::Sendmail, was not changed accordingly.


jeremy:
Thanks @jethrogb! Backported to 2.5.x and 2.6.x also.

PR #709, opened 2014-05-16
2.5.4 stable - NilClass is being extended when it should not be

pnomolos:
I'm not sure whether this is an issue with the Treetop grammar or the Treetop parser itself.  I tried to dig in a little but unfortunately I don't know enough about Treetop to fully make sense of what's going on here.

From lib/mail/parsers/rfc2822.rb, lines 2882-2893:

``` ruby
      i0 = index
      r1 = _nt_group
      r1.extend(Address0)
      if r1
        r0 = r1
      else
        r2 = _nt_mailbox
        if r2
          r0 = r2
        else
          @index = i0
          r0 = nil
        end
      end
```

The `r1` variable extends `Address0` before checking for truthiness, which results in `NilClass` being extended.

### DIFF
--- a/spec/mail/elements/address_list_spec.rb
+++ b/spec/mail/elements/address_list_spec.rb
@@ -24,6 +24,12 @@ describe Mail::AddressList do
       doing { Mail::AddressList.new(parse_text) }.should raise_error
     end
 
+    it "should not extend NilClass" do
+      parse_text  = 'test@lindsaar.net'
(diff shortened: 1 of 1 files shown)


jeremy:
Turns out this is a bug in Treetop and more recent versions don't fix it. Since we've switched parsers, we can't reasonably fix this in an old stable release.

PR #719, opened 2014-06-04
Don't strip out content that looks like the field name from the middle of a field

kjg:


kjg:
Is there anything holding this back from getting merged?

kjg:
This has been sitting for 9 months. I'd love to get some closure. Is this an acceptable change? If not, is there anything I can do to fix it up? Thanks!

If you all are swamped and are interested in having some help, I'm willing and available to help out with PRs.

corlissc:
@bf4, RFC-2822 would support the change that kjg made.  Depending on how the library splits up the original (raw) message header, the Regex would use either a start of string (\A) or a start of line (^).  With proper unfolding of the header lines, the best option would be a start of string (\A) as kjg has committed.

Please refer to RFC-2822 section 2.2 for the information that would be pertinent to the change being requested (https://www.ietf.org/rfc/rfc2822.txt).

bf4:
cc @jeremy @mikel @grosser @jeremyevans

jeremyevans:
I agree that the current behavior (stripping anywhere in the string) is wrong and the proposed behavior (only stripping at the start of the string) is better.

kjg:
@bf4 @jeremy I just rebased this to remove the conflict in the README, it sounds like this is ready to be merged now. Thanks!

grosser:
this still strips `Subject: Subject: Foobar`-> I think a better fix is doing less work / dirty hacks -> https://github.com/mikel/mail/pull/766

kjg:
@grosser I fully support #766 as a long term solution, but it sounds like it needs more work to address the API change by adding deprecation and what not. This is at least an improvement over what's currently released  and it's ready to go today and has support from others. 

Is there anything I can do to help you  get #766 into a state that the project will accept?

grosser:
I so far did not see anything breaking from this 'api change', I don't
expect anyone to use this afaik internal apis.
I'm fine with merging this, just sad that this is doing a shady change to
avoid solving the real problem :(

kjg:
What is shady about this change?

grosser:
It's still doing the wrong thing, just being more elaborate about it :D

kjg:
I'm not sure what's more elaborate about only matching at the beginning of the string, but I'm in full support of #766. Unfortunately it does seem like they are willing to accept #766 as is. If you'd like help getting it to a place that the maintainers are willing to accept, i'm very willing to help out.

This change at least makes the behavior more correct that it was before, and it's a pretty simple change to merge.

grosser:
Agreed, this is better than before and should not break anything.
No idea what is missing in 766 though ...

PR #723, opened 2014-06-05
add option to enable starttls for imap retriever

doits:
Anything else you need for a merge? Are there some docs or tests to update? Please point me to them and I'll do it if required.

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -2,6 +2,8 @@
 
 Features:
 
+* Add starttls support for imap retriever (doits)
+
 Performance:
 
 Bugs:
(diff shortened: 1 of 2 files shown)


krisdigital:
Until this option is merged, chances are that you can use imaps via port 993, then no STARTTLS is needed.

http://wiki.dovecot.org/SSL:
> Originally SSL support was added to protocols by giving them a separate "SSL port" (imaps, pop3s, etc.), where the SSL handshake starts immediately when client connects, and only after the session is encrypted the regular protocol handling begins. Using two separate ports for plaintext and SSL connections was thought to be wasteful and adds complexity for clients which may wish to make use of SSL when it is advertised, so STARTTLS command was added and intended to deprecate the SSL ports. Clients using STARTTLS work by connecting to the regular unencrypted port and immediately issue a STARTTLS command, after which the session is encrypted. After SSL handshake there is no difference between SSL port initiated connections and STARTTLS initiated connections.

````
retriever_method :imap, :address    => "mail.your-mail-server",
                              :user_name  => 'your user name',
                              :password   => '***************',
                              :enable_ssl => true,
                              :port => 993
````

Issue #618, opened 2013-09-02
ISO-8859-1 encoded body decoded as ascii/binary


the8472:
``` ruby
irb(main):077:0> part
=> #<Mail::Part:92631920, Multipart: false, Headers: <Content-Type: text/html; charset=ISO-8859-1>, <Content-Transfer-Encoding: quoted-printable>>
irb(main):078:0> part.charset
=> "ISO-8859-1"
irb(main):079:0> part.body.decoded.encoding
=> #<Encoding:ASCII-8BIT>
irb(main):091:0> part.body.raw_source[178..200]
=> "Vielen Dank f=FCr Ihren"
irb(main):103:0> part.body.decoded[164..184]
=> "Vielen Dank f\xFCr Ihren"
irb(main):113:0> part.body.decoded.encode('UTF-8', 'iso-8859-1')[164..184]
=> "Vielen Dank für Ihren"
```

PR #738, opened 2014-07-18
Set the Ruby string encoding from the charset

mkdynamic:
Related: https://github.com/mikel/mail/issues/618#issuecomment-49389043

When getting the decoded string back for the body of a `Message` which is text (i.e. `text?` is true), the original encoding (which should be present in the charset) is lost, and it currently always forces to UTF-8. That seems bad, since if we know the "right" encoding (which we do, presuming the charset on the message is correct) if seems more desirable to use it.

The encoding stuff seems super hairy, so if I've totally misunderstood something, let me know :)

Only implemented for Ruby 1.9+, I'm not sure it's possible for < 1.9 since we do not know the current string encoding to pass to `Iconv`. Let me know if there is a way.

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -2149,7 +2149,6 @@ module Mail
         else
           if encoding = Encoding.find(charset) rescue nil
             body_text.force_encoding(encoding)
-            return body_text.encode(Encoding::UTF_8, :undef => :replace, :invalid => :replace, :replace => '')
           end
         end
       end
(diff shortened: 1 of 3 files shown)


jeremy:
Good instincts, but for most email handling we want to work with a uniform encoding, regardless of incoming charset, much like `Encoding.default_external` behaves for files, databases, and other IO.

There's certainly room to make this more explicit: support something like a `Message#default_external` encoding that falls back to `Mail.default_external_encoding` and `Encoding.default_external`. If it's `nil`, preserve the internal charset encodings rather than converting to UTF-8.

jeremy:
Related: #558

PR #753, opened 2014-07-23
fix simple trailing whitespace

### LINKED ISSUE #742: Quoted printable keeps trailing =
since ruby does not want to fix this, mail should ...
https://bugs.ruby-lang.org/issues/7175

there already seems to be code for this in q_value_decode, just on the wrong line ...

fix:

``` Ruby
module DecodeWithTrailing
  def decode(str)
    super(str.sub(/=($|\\n)/, ""))
  end
end
class << Mail::Encodings::QuotedPrintable
  prepend DecodeWithTrailing
end

```

grosser:
fixes #742

https://bugs.ruby-lang.org/issues/7175

@bf4 @jeremy

### DIFF
--- a/lib/mail/encodings/quoted_printable.rb
+++ b/lib/mail/encodings/quoted_printable.rb
@@ -5,7 +5,7 @@ module Mail
   module Encodings
     class QuotedPrintable < SevenBit
       NAME='quoted-printable'
-   
+
       PRIORITY = 2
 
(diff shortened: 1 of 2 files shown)


jeremy:
Not pursuing this given that we're no longer stripping the final newline which resulted in the invalid trailing `=` in the first place.

PR #754, opened 2014-07-23
Forcing evaluation to integer prevents infinite looping with ruby-units

cbillen:
If you use ruby-units in your project the division of 1/2 will return a fraction, in turn this will yield an infinite loop in this method causing any call to mail to fail and freeze the system.  By simply ensuring we're casting to an integer we avoid any potential conflicts.


jeremy:
Fixed by #795

PR #766, opened 2014-07-31
remove duplicate field stripping

grosser:
@bf4 @jeremy @mikel 

calling afield with it's own name seems like a pretty useless feature, removing this breaks ~30 specs, that I'd fix if we decide to go this way ...

https://github.com/mikel/mail/pull/720

Ideally the .to_s would not be necessary, but something in mail or action_mailer passes in a class for content_transfer_encoding :/

fix:

``` Ruby
# https://github.com/mikel/mail/pull/766
Mail::CommonField.class_eval do
  def strip_field(field_name, value)
    value.to_s
  end
end
```

### DIFF
--- a/spec/mail/fields/common/common_field_spec.rb
+++ b/spec/mail/fields/common/common_field_spec.rb
@@ -4,7 +4,7 @@ require 'spec_helper'
 describe Mail::CommonField do
 
   describe "multi-charset support" do
-    
+
     before(:each) do
       @original = $KCODE if RUBY_VERSION < '1.9'
     end
@@ -16,13 +16,13 @@ describe Mail::CommonField do
     it "should return '' on to_s if there is no value" do
       expect(Mail::SubjectField.new(nil).to_s).to eq ''
     end
-    
+
     it "should leave ascii alone" do
       field = Mail::SubjectField.new("This is a test")
       expect(field.encoded).to eq "Subject: This is a test\r\n"
       expect(field.decoded).to eq "This is a test"
     end
-    
+
     it "should encode a utf-8 string as utf-8 quoted printable" do
       value = "かきくけこ"
       if RUBY_VERSION < '1.9'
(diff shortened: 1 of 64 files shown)


grosser:
ran into this gain when doing outbound email ... would love to get this merged ... I don't see anyone using this api anyway ...

grosser:
@jeremy rebased, can we get this merged now ?

(using this patch in apps that both handle inbound and outbound mail without any issues since over a year now ... )

kjg:
@grosser You might have to rebase this again

grosser:
done!

spec/mail/fields/common/common_field_spec.rb
```diff
@@ -69,17 +69,8 @@
...
   end
 
-  context "when including the field name" do
-    it "does not strip out content that looks identitcal to the field name" do
-      field = Mail::SubjectField.new("Subject: Subject: for your approval")
-      expect(field.decoded).to eq("Subject: for your approval")
-    end
-  end
-
-  context "when not including the field name" do
```
kjg: Can we leave this spec in place?

grosser: feels useless to me ... but added it back if that makes you happy ...

kjg: Thanks! It's to prove that this bug is still fixed: https://github.com/mikel/mail/pull/719/files

jeremy:
#1111

seb-sykio:
since I updated to 2.7.0, I have this issue :

>       Failure/Error: mm = Dkim.sign(mm, :domain => xxx, :selector => 'mail', :private_key => OpenSSL::PKey::RSA.new(xxx))
>       
>       NoMethodError:
>         undefined method `strip_field' for #<Mail::DkimField:0x0000564a55a59bf0 @charset="utf-8">
>       # /xxx/.rvm/gems/ruby-2.4.2@xxx/gems/dkim-1.0.0/lib/mail/dkim_field.rb:12:in `initialize'

PR #782, opened 2014-08-17
Use composition over inheritance for `PartsList`

robin850:
Hello,

This pull request makes this gem compatible with Rubinius relying on the composition over inheritance technique. Even Rails no longer uses the inheritance pattern (see [this article](http://tenderlovemaking.com/2014/06/02/yagni-methods-are-killing-me.html) and rails/rails#16299).

For now, there is a compatibility layer that delegates any missing methods to the parts list's `@parts` so it behaves like an array. This is up to you but you may need to deprecate this path and tell people to access `#parts` first, then call the given method or provide your very own API (through `Enumerable` for instance).

If you want me to squash some commits together let me know !

Refs rubinius/rubinius#3050.

This has been tested under MRI 2.1.2 and 1.9.3, JRuby 1.7.13 and Rubinius 2.2.10.

Have a nice day.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,5 +41,4 @@ matrix:
   allow_failures:
     - rvm: ruby-head    # green, but unstable
     - rvm: jruby-head   # green, but unstable
-    - rvm: rbx-2        # see https://github.com/rubinius/rubinius/issues/3050
   fast_finish: true
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -1,5 +1,6 @@
 == HEAD
 
+* Make the gem compatible with Rubinius (Robin Dupret)
 * #808 - Mail::Field correctly responds_to? the methods of its instantiated field (thegcat)
 * #772 - normalize encoding matchers (grosser)
 * #789 - Fix encoding collapsing not dealing with multiple encodings in 1 line (grosser)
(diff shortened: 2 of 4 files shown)


robin850:
Ok, sorry it took so long to get back to you but the feedback seems to be addressed. :-) @bf4 : Thanks a ton for working on making RSpec more compliant with Rubinius ; hoping the pull request will be grabbed soon ! :heart:

Do you want me to put back Rubinius in the list of allowed failures ? At least the JRuby build is broken as well due to travis-ci/travis-ci#3067.

> it seems that represent_object isn't available for < 1.9.3

Is it really a problem ? In Ruby 1.9.2 and less, the default YAML engine is Syck so the `encode_with` method won't ever be called.

Let me know if there's anything missing! Happy new year to you guys! :tada:

bf4:
Trying this with RSpec head https://travis-ci.org/mikel/mail/builds/50076945

bjfish:
@robin850 This PR built successfully for me locally with rubinius, maybe reopen the PR to trigger a new build?

robin850:
It looks like you're right @bjfish, everything is green now ! \o/

@bf4 : Is there anything missing before shipping it please ? :-)

bf4:
Looks good to me @jeremy @mikel

PR #790, opened 2014-09-07
Mail::Field responds_to? delegated methods properly

owenr:
Mail::Field delegates method calls to the contained field using `method_missing` but did not also override `respond_to?`. As an example, attempts to call `mail[:to].try(:decoded)` always return `nil` instead of the decoded field value.
`respond_to?` used instead of `respond_to_missing?` for Ruby 1.8.7 compatibility. (See http://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding)
Tests pass on MRI 1.8.7-p375, 1.9.3-p547, 2.0.0-p247.

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -9,6 +9,7 @@ Performance:
 
 Bugs:
 * #736 - Mail.new copes with non-UTF8 messages marked as UTF8 (jeremy)
+* Mail::Field responds_to? delegated methods properly (owenr)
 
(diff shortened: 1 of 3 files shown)


owenr:
I see this was reimplemented two months later in #836 (addressing #808). Consider taking the additional test from my pull request, and let me know if there is some reason why this request wasn't accepted so I can improve the quality of future submissions.

grosser:
sadly just this repo not getting any love / not having dedicated maintainers ...

jeremy:
Sorry about that, @owenr! Missed your earlier implementation during issue triage. Thank you 😊

PR #792, opened 2014-09-15
Allow blank filenames in Content-Disposition field

robinroestenburg:
Attachments containing an empty string as a filename in the Content-Disposition header are skipped.

This was introduced as a side-effect of commit https://github.com/mikel/mail/commit/b1b17961dbc4adbf1205b5201495b2fa6f4afebd.

Take, for example, the following Content-Disposition header:

```
Content-Disposition: inline;
  filename=""
```

Prior to that commit the parameters in the ContentDispositionField class would return:

``` ruby
{"filename"=>"\"\""}
```

After the commit the following is returned:

``` ruby
{"filename"=>""}
```

This is correct, but the code returning the filename from the ContentDispositionField class checks on `!blank?`. Which would return true before, and now returns false.

This would result in `Message#find_attachment` not decoding the attachment because nil instead of "" is returned.

For this PR I made the `ContentDispositionField#filename` implementation match the one in `ContentTypeField#filename`.

### DIFF
--- a/lib/mail/fields/content_disposition_field.rb
+++ b/lib/mail/fields/content_disposition_field.rb
@@ -3,10 +3,10 @@ require 'mail/fields/common/parameter_hash'
 
 module Mail
   class ContentDispositionField < StructuredField
-    
+
     FIELD_NAME = 'content-disposition'
     CAPITALIZED_FIELD = 'Content-Disposition'
-    
+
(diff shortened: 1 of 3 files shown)


grosser:
thx for merging/fixing all these issues @jeremy ! ❤️

Issue #279, opened 2011-09-02
Mail::Address accepts invalid emails


jasonmp85:
The [grammar specified in RFC 5322§3.2.3](http://tools.ietf.org/html/rfc5322#section-3.2.3) should not accept emails with dots appearing at the very beginning of the local part of the address. Additionally, local parts with two or more consecutive periods should also be rejected.

Or in the more succinct words of [Wikipedia](http://en.wikipedia.org/wiki/Email_address#Syntax):

> The local-part of the email address may use any of these ASCII characters RFC 5322 Section 3.2.3:
> - Uppercase and lowercase English letters (`a`–`z`, `A`–`Z`) (ASCII: 65-90, 97-122)
> - Digits `0` to `9` (ASCII: 48-57)
> - Characters `!` `#` `$` `%` `&` `'` `*` `+` `-` `/` `=` `?` `^` `_` ``` `{` `|` `}` `~` (ASCII: 33, 35-39, 42, 43, 45, 47, 61, 63, 94-96, 123-126)
> - Character `.` (dot, period, full stop) **provided that it is not the first or last character, and provided also that it does not appear two or more times consecutively** (e.g. `John..Doe@example.com`).

(Emphasis mine)

While the mail gem's grammar correctly rejects local parts with a dot as the final character, it incorrectly accepts a dot in the initial position and local parts with sequences of more than two dots.

PR #797, opened 2014-09-21
don't parse addresses with invalid domains

hengwoon:
Per RFC 2822, domains like ".domain.com" or "domain..com" are not valid.

```
# RFC 2822
dot-atom        =       [CFWS] dot-atom-text [CFWS]
dot-atom-text   =       1*atext *("." 1*atext)
```

Domains should be a domain_text element followed by zero or more
domain_text elements prefixed with "."


kennethkalmer:
:+1: on this!

bpot:
:+1: nice catch! I can't remember any reason why this deviated from the spec.

bf4:
@jeremy ok to merge?

jeremy:
Previously reported in #279 as a deviation from RFC, patched in #286, and reverted in #397. These deviations are not-uncommonly seen in the wild, so we purposefully support parsing them.

PR #802, opened 2014-09-27
Decoding adjustments

mreinsch:
I've rebased some of my old encoding related pull requests on master. Thanks to @grosser and @jeremy  some of this stuff is finally getting into mail :)

Part of this pull request are:
- some more tests based on some strange stuff we've seen that broke the mail lib
- adjusted the b_value_decode to return the original string instead of some part of it (str is overwritten by the match).
- Japanese users love to use emoji which are sometimes non-defined UTF-8 characters. So telling the encode method to simply skip those gives us at least a decoded string without the characters that can't be displayed anyway. Better than the encoded version.

Please consider merging this, we've been using those since quite a while and I'd really appreciate it if I could get rid of my patch branch. Thanks!

### DIFF
--- a/lib/mail/version_specific/ruby_1_9.rb
+++ b/lib/mail/version_specific/ruby_1_9.rb
@@ -78,6 +78,8 @@ module Mail
       [Ruby19.encode_base64(str), encoding]
     end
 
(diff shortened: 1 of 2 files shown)


jeremy:
Test cases ~passing on master 😊

PR #806, opened 2014-09-30
Stop quote_phrase from modifying input string encoding

c2c533c4f237:
In the case where it acted to quote the input string, quote_phrase
shouldn't modify the encoding of the input string to be ASCII-8BIT, but
should leave the encoding as it found it.


jeremy:
Fixed by #974.

Issue #812, opened 2014-10-14
Memory Use exploded in 2.6


schneems:
Rails 4.2 is moving to Mail 2.6.1 and i'm seeing a huge increase in memory usage in my running application. After some digging, it looks like the culprit was the Mail version bump.

To demonstrate the issue, here's a script to reproduce without rails:
### Build Gemfiles

```
$ mkdir 2.6.1
$ cat <<  EOF > 2.6.1/Gemfile
  source 'https://rubygems.org'
  gem 'mail', '2.6.1'
  gem 'get_process_mem'
EOF

$ mkdir 2.5.4
$ cat <<  EOF > 2.5.4/Gemfile
  source 'https://rubygems.org'
  gem 'mail', '2.5.4'
  gem 'get_process_mem'
EOF
```
### Install

```
$ env BUNDLE_GEMFILE=2.6.1/gemfile bundle install
$ env BUNDLE_GEMFILE=2.5.4/gemfile bundle install
```
### Test script

```
$ cat <<  EOF > script.rb
  require 'get_process_mem'
  mem = GetProcessMem.new
  before = mem.mb
  require 'mail'
  after = mem.mb
  puts after - before
EOF 
```
### Results

```
$ env BUNDLE_GEMFILE=2.6.1/gemfile bundle exec ruby script.rb
# => 41.44140625 # mb
$ env BUNDLE_GEMFILE=2.5.4/gemfile bundle exec ruby script.rb
# => 11.41015625 # mb
```
## EWAT

Mail 2.6.1 takes up almost 4 times more memory at Require time 0_o. Initial guess is that this has to do with the switch from treetop to ragel via @bpot 

cc/@matthewd

vishnurkrishnan:
How can we do this with rails 3.2.22 and ruby 2.0, because i can't upgrade the mail gem to 2.6, it seems that is not compatible with action mailer, i was following this https://www.schneems.com/2014/11/07/i-ram-what-i-ram.html

grosser:
use a money-patched version that loosens the dependency
gem "mail",                 git: 'https://github.com/zendesk/mail', ref:
'261-as-259' # fake 2.5.999 version so we can use with actionmailer 3

vishnurkrishnan:
Yes, but here in this https://www.schneems.com/2014/11/07/i-ram-what-i-ram.html says need to upgrade to 2.6.3 or higher, will it be possible with this patch?

grosser:
[...]

PR #815, opened 2014-10-22
autoload parsers for load-time speed and memory usage

grosser:
https://github.com/mikel/mail/issues/812

@bf4 
### Risks
- None

### DIFF
--- a/lib/mail.rb
+++ b/lib/mail.rb
@@ -76,7 +76,7 @@ module Mail # :doc:
 
   require 'mail/envelope'
 
-  require 'mail/parsers'
+  register_autoload :Parsers, "mail/parsers"
 
   # Autoload header field elements and transfer encodings.
   require 'mail/elements'


tenderlove:
If this doesn't load parsers until runtime, and my app needs the parser, does that mean that the lazy loading will ruin CoW advantages of eager loading?  I mean, should people who use the parser and forking severs know that they should eagerly load?  (Otherwise this could possibly increase memory usage)

SamSaffron:
aren't we usually doing mail parsing in sidekiqs anyway?

grosser:
Yes balancing dev vs production needs is always problematic,
in rails you should use eager_load in production and maybe something
similar in other frameworks.

We do mail parsing in resque with forks.

tenderlove:
> in rails you should use eager_load in production and maybe something
> similar in other frameworks.

`eager_load` doesn't guarantee that the mail parser constant will be touched (e.g. the first time the constant is referenced could be in a method that isn't executed until runtime). :-/

grosser:
but adding it to eager_load_paths would do the trick ?

PR #816, opened 2014-10-22
Remove creation of ActiveSupport module

jeremyevans:
If ActiveSupport is not already defined, mail should not attempt
to define it.  Otherwise it can break existing code that checks
if the ActiveSupport module is defined and assumes it is loaded
if so.

Here's an example:
https://github.com/icalendar/icalendar/blob/0459def239298c40d6e08f70bc6e641405e5298e/lib/icalendar/values/time_with_zone.rb#L15

If you don't have active_support installed, and you load the mail
and icalendar gems, icalendar ends up broken.


rpanachi:
Why this have not yet been merged? Can I help someway?

PR #818, opened 2014-10-23
Download spam corpus and verify all in a script

bf4:
For consideration

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@
 /.rbenv-vars
 /.ruby-version
 gemfiles/*.lock
+corpus/spam
--- a/corpus.bash
+++ b/corpus.bash
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+# Usage: bash corpus.bash
+# Downloads and unpacks the spam corpus, if not present
+# Runs rake corpus:verify_all && exit 0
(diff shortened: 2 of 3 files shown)


bf4:
@mikel @jeremy any objections?

jeremy:
What's your thinking here @bf4 ? Can see this is making it easier to download the spam corpus, but not sure about the significance of introducing the test script.

If it were for convenience alone, I'd expect a couple of new rake tasks, e.g. `corpus:spam:download` and `corpus:spam:verify`

bf4:
I suppose I could see about moving the code from bash to Ruby.

My thinking is basically just to make it easier to run the test.

bf4:
Also, it was part of my interest in writing up a 'how to release mail' that I wanted to extract from @mikel  :)

jeremy:
Cool. Using rake tasks means you can represent the dependencies that'd otherwise need imperative code (like whether to download, how to unpack, etc). Makefile!

bf4:
I rebased off of master and started converting the script to ruby/rake... sharing it here just in case you want to look at it between now and when I come back to this.

jeremy:
Closing this out until we wrap it up. Moved the branch off this origin, reserved for release branches, and pushed an update to https://github.com/jeremy/mail/tree/verify_corpus 👍

Issue #467, opened 2012-11-22
Attachments with non-ASCII filenames are not decoded.


lawrencepit:
Use case: when you send an email with Outlook with an attachment filename using e.g. Japanse characters then the attachment.filename method does not return the filename decoded if the filename is based on the content-disposition header.

It appears this bit from RFC 2231 is not supported by Mail:

```
   Asterisks ("*") are reused to provide the indicator that language and
   character set information is present and encoding is being used. A
   single quote ("'") is used to delimit the character set and language
   information at the beginning of the parameter value. Percent signs
   ("%") are used as the encoding flag, which agrees with RFC 2047.
```

In `spec/fixtures/emails/attachment_emails/attachment_with_quoted_filename.eml` there is an example of such an encoded filename:

```
Content-Disposition: inline;
	filename*=ISO-8859-1''Eelanal%FC%FCsi%20p%E4ring.jpg
```

See for more examples / tests:
http://greenbytes.de/tech/tc2231/#encoding-2231-char

jeremy:
#819

PR #819, opened 2014-10-24
Fixes for #467

okkez:
Decode string in Content-disposition header parameter such as "filename" properly.


kjg:
It looks like #998 (https://github.com/mikel/mail/commit/6fa8179c3806d52ff9ef8b308fe9c4dbbb4ff3da) fixed the blank charset part of this PR and #1000 is the fix for the transcode part.

PR #824, opened 2014-10-31
fix sorting

venkatesh3007:
mid = (lo + hi) / 2 return 1/2. This causes an infinite while loop.


jeremy:
Duplicate of #754

jeremy:
Fixed by #795

PR #826, opened 2014-11-02
Remove redundant parse in address fields

bpot:
The value is already parsed by the call to `super`. The [`StructuredField` initializer](https://github.com/mikel/mail/blob/17457e44596308d8092d340e372fb8a582bb66be/lib/mail/fields/structured_field.rb#L29) calls the [`value` setter](https://github.com/mikel/mail/blob/17457e44596308d8092d340e372fb8a582bb66be/lib/mail/fields/common/common_address.rb#L96) which parses the value.

I dug around to see if there was a reason for this but it just looks like the duplication was unintentionally added in edbe7861e75721f376f53dd51ca1a68c80a3573c

### DIFF
--- a/lib/mail/fields/bcc_field.rb
+++ b/lib/mail/fields/bcc_field.rb
@@ -39,7 +39,6 @@ module Mail
     def initialize(value = '', charset = 'utf-8')
       @charset = charset
       super(CAPITALIZED_FIELD, strip_field(FIELD_NAME, value), charset)
-      self.parse
       self
     end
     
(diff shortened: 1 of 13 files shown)


kjg:
Is there anything I can do to help this get merged in?

kjg:
This looks pretty straightforward; all the tests still pass. I'd love to do some refactoring in some of these field classes and I'd love to see this change merged in first.

kjg:
This looks pretty straightforward; all the tests still pass. I'd love to do some refactoring in some of these field classes and I'd love to see this change merged in first.

grosser:
@jeremy can we have a review / merge ?

PR #829, opened 2014-11-14
Reduce memory usage by 2/3 by using cut down mime type implementation

jeremyevans:
Mail uses the mime/types gem, but uses only a tiny portion of it.
Unfortunately, the mime/types gem has a huge memory footprint.  This
replaces the use of mime/types with an internal library that implements
only the parts of mime/types that mail uses, which are:

* Determining whether a content type is binary or not, to allow
  guess_encoding to determine 7bit or binary.
* Guessnig the content type based on a filename, if a content type is
  not explicitly specified.

This adds a copy_mime rake task, which will use the mime/types gem and
pull out only the information needed for the parts that mail uses, and
stores that cache in lib/mail/mime_type_cache.rb.  This way if
mime/types adds types, it's easy to rebuild the cache.

To limit the amount of memory used, the mime types are stored as one of
the following two classes: Mail::MimeType (if 7bit) or
Mail::MimeType::Binary (if binary).  Each just stores the content type
as a string.  Separate classes are used so that each instance doesn't
have to store whether it is binary or not, which saves memory.

This is currently just a proof of concept to get feedback, more
documentation should be added before merging, at the least.

The 2/3 reduction in memory was taken from these numbers:

ruby RSS in KB:
without require mail:       6204
with require mail, before: 28860
with require mail, after:  14452


jeremyevans:
The large memory usage issue is addressed by mime-types 2.6 with the columnar store.  I'll send a pull request to use that.

PR #830, opened 2014-11-15
Remove ActiveSupport definition dependency; Update to Unicode 7.0.0

bf4:
Closes #816


edit-christian:
Just ran into this problem with mail in a sinatra project, which does not use activesupport. Definition of ActiveSupport in mail (unicode.rb) does cause problems with thirdparty gems.

I don't want to include ActiveSupport in my project, so which would be my best approach at the moment?
Any plans of merging this fix with the master?

bf4:
@jeremy I just rebased this against master (though there were no merge conflicts 934f02f85cab14f415208fd573be093dea23b3bd )

bf4:
looks like unicode 8.0 is now released

thhermansen:
It would be really nice if this got merged :)

lib/mail/multibyte/unicode.rb
```diff
@@ -390,11 +402,3 @@ def database
...
   end
 end
-
-unless defined?(ActiveSupport)
-  module ActiveSupport
-    unless const_defined?(:Multibyte)
-      Multibyte = Mail::Multibyte
-    end
-  end
-end
```
jeremy: Do we have anything that depends on this shim?

thhermansen: These lines of code seems to have been added here: b9fd65b3561b0c1b3a2fca502bbfe1908e74591b. The intent of that commit was "unicode dat file is a marshal of some ActiveSupport". This pull request removes that dependency by building the dat file on our own, without ties to ActiveSupport.

I'm no expert in neither this mail gem nor encoding, but to me this seems like the solution which should have been done three years ago.

jeremy: Nice – Got it!

thhermansen:
Thank you!

bf4:
:rainbow:

PR #834, opened 2014-11-19
try to load AS instead of seeing if it is loaded to avoid overwriting mb_chars

### LINKED ISSUE #833: mb_chars collides with ActiveSupport mb_chars
we should either use some clean wrapper or name the method mail_mb_chars

thoughts ? @jeremy @bf4

grosser:
@bf4 @jeremy 

fixes #833


bf4:
@grosser are you still using this patch?

grosser:
I fixed it in our code by using the right require order, so I don't really care about this PR ...

PR #839, opened 2014-11-25
Allow more options when using Message#add_file

jeremyevans:
Previously, Message#add_file would ignore options other than
:filename and :content.  This breaks the principal of least
surprise.  The documentation for #add_file says "see also
attachments", and attachments shows other possible options,
but those are not supported.

This changes the code to pass an options hash given to add_file
through to attachments[]=, allowing you to use options such as
:mime_type or :content_disposition.  Currently, you have to do
the following:

  add_file :filename=>"cal.ics", :content=>c.to_ical
  attachments.last.content_type 'text/calendar'

You would think you could work around this using:

  attachments["cal.ics"] = {:content=>c.to_ical,
                            :mime_type=>'text/calendar'}

However, that doesn't work because unlike add_file,
attachments[]= doesn't automatically convert the message to
multipart, which results in the removal of the existing body.
With this change, you can now do:

  add_file :filename=>"cal.ics", :content=>c.to_ical,
           :mime_type=>'text/calendar'

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -1772,7 +1772,10 @@ module Mail
         filedata = File.open(values, 'rb') { |f| f.read }
       else
         basename = values[:filename]
-        filedata = values[:content] || File.open(values[:filename], 'rb') { |f| f.read }
+        filedata = values
+        unless filedata[:content]
+          filedata = values.merge(:content=>File.open(values[:filename], 'rb') { |f| f.read })
+        end
       end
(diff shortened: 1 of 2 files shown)


jeremyevans:
It's been over 3 months since this pull request was opened.  This doesn't break anything and adds a useful feature. Any chance this could be considered?  Are there any problems with this that you would like me to fix?

jeremy:
:+1:

PR #840, opened 2014-12-02
*.msg attachments sent from Outlook are not detected as attachments

pavel-d:
I've encountered an issue when processing email which was sent from Outlook 2007 and has *.msg attachment in it.

You can see full raw message in my commit, here is a part of it:

```
------=_NextPart_000_0EE8_01D00D91.AB93BEF0
Content-Type: message/rfc822
Content-Disposition: attachment

...
```

Attachments that doesn't contain filename in the Content-Disposition header are skipped. However any other email client (Gmail, Mac OS Mail, Outlook) processes them correctly and just assigns some random name to them.

### DIFF
--- a/lib/mail/attachments_list.rb
+++ b/lib/mail/attachments_list.rb
@@ -5,7 +5,7 @@ module Mail
       @parts_list = parts_list
       @content_disposition_type = 'attachment'
       parts_list.map { |p|
-        if p.content_type == "message/rfc822"
+        if p.content_type == "message/rfc822" && p.content_disposition != 'attachment'
           Mail.new(p.body).attachments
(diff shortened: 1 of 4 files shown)


pavel-d:
@mikel, have you had a chance to look at it?

craig1410:
Hi, I've run into this issue as well @mikel and was wondering whether you think @pavel-d 's solution is acceptable? It looks good to me certainly.

jeremy:
We never treat rfc822 parts as attached files since we recurse *through* them looking for additional attached files. To treat rfc822 parts as attachments, we need to change behavior and stop recursing through them. Perhaps for 3.0.

PR #841, opened 2014-12-04
Fix address with Q-encoded display name containing double quotes

rafbm:
I added a spec for parsing the following address:

```
=?ISO-8859-1?Q?"Jan_Kr=FCtisch"?= <jan@krutisch.de>
```

Before the fix, display name was parsed as `Jan_Kr=FCtisch` because when a quoted string was found (ie. `"Jan_Kr=FCtisch"`), it was used as the display name and the surrounding characters were ignored.

After the fix, display name is properly encoded as `Jan Krütisch`. Code-wise, the `qstr` variable is now used as the display name instead of `phrase` only if `qstr` is equal to `phrase` minus the quotes.

I would have liked to avoid creating the `%("#{qstr}")` string, but I did not see another way. Other than that performance/memory should not be affected.

### DIFF
--- a/lib/mail/parsers/address_lists_parser.rb
+++ b/lib/mail/parsers/address_lists_parser.rb
@@ -91,7 +91,7 @@ module Mail::Parsers
         when :angle_addr_s
           if phrase_e
             phrase = s[phrase_s..phrase_e].strip
-            if phrase[0] == DOUBLE_QUOTE && phrase[-1] == DOUBLE_QUOTE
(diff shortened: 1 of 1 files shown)


rafBM:
Sorry for the delay!

Yes I am confident enough with the specs I just added for comments inside Q-encoded display names.

rafBM:
Mmh not sure why specs fail on Travis. All green on my machine. Can someone have a look?

PR #849, opened 2015-01-27
Handle parts with an un-parsable content-disposition

kjg:
Otherwise a NoMethodError is raised when calling `Part#inline?`


jeremy:
:+1: nice catch.

Could you squash the commits together?

kjg:
No problem. Squashed.

kjg:
FYI, the travis failure appears to be due to a network failure In just one job of the test matrix.

kjg:
Can this get merged now or is there something else I can do to get it acceptable?

Issue #706, opened 2014-05-10
Attachments in mail messages with both html_part and text_part don't show up in Yahoo! or AOL webmail.


lighth7015:
As per this [StackOverflow post](http://stackoverflow.com/questions/8217213/html-emails-missing-attachments), I've been running into this myself, and its got me scratching my head as to why.

I'm running **ruby 1.9.3p125 (2012-02-16 revision 34643) [x86_64-freebsd7.0]**, with mail version **2.5.4**.

jeremy:
#853

PR #853, opened 2015-02-06
Allow overwriting Mail::Message sort order… for real

rafbm:
So I tried to build the following email:

```rb
mail = Mail.new do
  # FIXME: This does not work.
  body.set_sort_order []

  part do |p|
    p.text_part { body 'Hello world!' }
    p.html_part { body 'Hello <b>world</b>!' }
  end

  attachments['README.txt'] = 'some text'
end
```

An alternative part (plain + HTML) followed by a .txt attachment.

The resulting email was structured this way:

```
multipart/mixed
  text/plain (attachment)
  multipart/alternative
    text/plain
    text/html
```

Starting with the attachment looked like a bug to me, so I dug a bit. I found about `Mail::Body#sort_parts!` and the default `@part_sort_order`, so I tried disabling the sorting altogether by passing `Mail::Body#set_sort_order` an empty array (see `FIXME` comment above).

It did not work, because `Mail::Message#ready_to_send!` does a hard-coded `parts.sort!` with the default order, effectively making `Mail::Body#set_sort_order` useless.

This patch contains the fix. The example above now produces the following structure:

```
multipart/mixed
  multipart/alternative
    text/plain
    text/html
  text/plain (attachment)
```

### DIFF
--- a/lib/mail/body.rb
+++ b/lib/mail/body.rb
@@ -31,7 +31,7 @@ module Mail
       @preamble = nil
       @epilogue = nil
       @charset  = nil
-      @part_sort_order = [ "text/plain", "text/enriched", "text/html" ]
+      @part_sort_order = [ "text/plain", "text/enriched", "text/html", "multipart/alternative" ]
       @parts = Mail::PartsList.new
       if string.blank?
         @raw_source = ''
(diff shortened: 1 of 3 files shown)


mpapis:
most likely related to:  #706 #675  #655

jeremy:
Squashed 7812a77aab2d24b19087e5df2df7fba2d82ea004

PR #856, opened 2015-02-19
Add Logger delivery method

zacholauson:
Tested against: 1.8.7, 1.9.3, 2.1.5, and 2.2.0

Here is how it looks when an email gets delivered:
screen shot 2015-02-20 at 2 47 01 pm

### DIFF
--- a/lib/mail/network/delivery_methods/stdout.rb
+++ b/lib/mail/network/delivery_methods/stdout.rb
@@ -1,0 +1,26 @@
+require 'mail/check_delivery_params'
+require 'logger'
+
+module Mail
+  class Stdout
+    include Mail::CheckDeliveryParams
+
+    def initialize(values)
+      self.settings = default_settings.merge!(values)
+    end
+
+    attr_accessor :settings
+
+    def deliver!(mail)
+      check_delivery_params(mail)
+      settings[:logger].send(settings[:severity], mail.encoded)
+    end
+
+    private
+
+    def default_settings
+      { :logger => ::Logger.new(STDOUT), :severity => :info }
+    end
+
+  end
+end
--- a/spec/mail/network/delivery_methods/stdout_spec.rb
+++ b/spec/mail/network/delivery_methods/stdout_spec.rb
@@ -1,0 +1,63 @@
+# encoding: utf-8
+require 'spec_helper'
+
+describe "Stdout Delivery Method" do
+
+  before(:each) do
+    # Reset all defaults back to original state
+    Mail.defaults do
+      delivery_method :smtp, { :address              => "localhost",
+                               :port                 => 25,
+                               :domain               => 'localhost.localdomain',
+                               :user_name            => nil,
+                               :password             => nil,
+                               :authentication       => nil,
+                               :enable_starttls_auto => true  }
+    end
+  end
+
+  let(:mail) do
+    Mail.new do
+      from    'dschrute@dm.com'
+      to      'mscarn@dm.com'
+      subject 'Beet Bandit'
+    end
+  end
+
+  it "sends an email to STDOUT with 'info' severity by default" do
+    Mail.defaults do
+      delivery_method :stdout
+    end
+
+    logger = mail.delivery_method.settings.fetch(:logger)
+
+    expect(logger).to receive(:info).with(mail.encoded)
+
+    mail.deliver!
+  end
+
+  it "can be configured with a custom logger" do
+    custom_logger = double("custom_logger")
+
+    Mail.defaults do
+      delivery_method :stdout, { :logger => custom_logger }
+    end
+
+    expect(custom_logger).to receive(:info).with(mail.encoded)
+
+    mail.deliver!
+  end
+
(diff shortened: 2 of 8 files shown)


carsonreinke:
@zacholauson Cool, but maybe this should use `Logger` instead and option to pass in.

zacholauson:
Good point @carsonreinke, not sure why I didn't think of using Logger. I'll update this later today.

bf4:
I don't think this is a feature in the gem that I'd want to maintain.  @jeremy @mikel ?

zacholauson:
@bf4 could you elaborate as to what you think would be difficult to maintain? Is there something you'd like to see changed in the code?

zacholauson:
Fixed merge conflicts.

bf4:
@jeremy @mikel @grosser @jeremyevans thoughts on having a STDOUT delivery method baked in, and the implementation?

jeremyevans:
I think a STDOUT or a logger delivery method is fine, but the current name is misleading.  This is currently a Logger delivery method, which just happens to have a default logger which logs to STDOUT. 

I think this facility should it should probably be renamed (`delivery_method  = :logger`), or the implementation be changed to something like `$stdout.puts mail.encoded`. Of the two, I prefer the former, as it is more flexible.

In terms of the implementation, this creates an unused Logger instances if the user passes in their own logger.  It would be best to try to avoid that.

spec/mail/network/delivery_methods/stdout_spec.rb
```diff
@@ -0,0 +1,63 @@
+# encoding: utf-8
+require 'spec_helper'
+
+describe "Stdout Delivery Method" do
+
+  before(:each) do
+    # Reset all defaults back to original state
```
grosser: looks more like 'change global state so that this test passes', whatever test unsets them should clean them up

lib/mail/network/delivery_methods/stdout.rb
```diff
@@ -0,0 +1,26 @@
...
+      check_delivery_params(mail)
+      settings[:logger].send(settings[:severity], mail.encoded)
+    end
+
+    private
+
+    def default_settings
+      { :logger => ::Logger.new(STDOUT), :severity => :info }
+    end
+
```
grosser: unnecessary newline

spec/mail/network/delivery_methods/stdout_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    Mail.defaults do
+      delivery_method :smtp, { :address              => "localhost",
```
grosser: If you 2-space indent everything there is less space wasted and nothing needs to change when the first line changes:
```Ruby
delivery_method :smtp, 
  :address              => "localhost",
  :port                 => 25,
  :domain               => 'localhost.localdomain',
  :user_name            => nil,
  :password             => nil,
  :authentication       => nil,
  :enable_starttls_auto => true
```

grosser:
In rails I am used to seeing the email in the logs during development, but I assume that's rails doing stuff and not mail itself ... :+1:

zacholauson:
> I think a STDOUT or a logger delivery method is fine, but the current name is misleading. This is currently a Logger delivery method, which just happens to have a default logger which logs to STDOUT.
>
> I think this facility should it should probably be renamed (delivery_method = :logger), or the implementation be changed to something like $stdout.puts mail.encoded. Of the two, I prefer the former, as it is more flexible.

I agree @jeremyevans, I'll update this to be Logger.

> In terms of the implementation, this creates an unused Logger instances if the user passes in their own logger. It would be best to try to avoid that.

I didn't even think about that, thanks for catching that, I'll fix it.

lib/mail/network/delivery_methods/stdout.rb
zacholauson: I thought this was convention, as I had seen it in many classes (ex: [Mail::Configuration](https://github.com/mikel/mail/blob/b159e0a542962fdd5e292a48cfffa560d7cf412e/lib/mail/configuration.rb#L72)), but looking back it doesn't appear to be consistent across all. I can remove it if you would prefer.

spec/mail/network/delivery_methods/stdout_spec.rb
zacholauson: :+1: I like that, updating now.

spec/mail/network/delivery_methods/stdout_spec.rb
zacholauson: Hmm, this is from every other delivery method spec except for Mail::SendMail. While it doesn't seem to be required to be in the logger spec file, if I remove it from all delivery method specs, I get intermittent failures. I'll have to look more into why this exists.

zacholauson:
Fixed merge conflicts.

zacholauson:
Fixed merge conflicts.

zacholauson:
Fixed merge conflicts.

jeremy:
Thanks @zacholauson!!

PR #278, opened 2011-08-27
Adds Mail::Message#without_attachments! that removes all message's attachments (issue #33)

vitorbaptista:
Hi,

I've implemented a method #without_attachments! that requires less memory than the solution proposed by ledermann in issue #33. I'm working on creating a #without_attachments that's not destructive, simply returning a new instance with the attachments removed, but haven't had success so far.

The only problem with this implementation, AFAIK, is that it doesn't works for RFC822 emails. I couldn't find out how to remove only te attachment from its body. Any ideas?

Thanks,
Vítor Baptista.

PR #858, opened 2015-03-02
Change Message#without_attachments! to delete attachments recursively

TylerRick:


jeremy:
Nice implementation. But why? What motivates this change?

jeremy:
Saw the motivation in #278. Fixed the immediate bug in #1301. Lining this up next.

ahmeij:
Is there something blocking this pull request? I am especially interested in `recursive_delete_if` method in parts list.

If there is something specific to be done I'd love to help.

barodeur:
Here is a small module I've been using as a workaround. I'll be very happy to submit something similar as a PR.

```ruby
module MailHelper
  module_function

  def delete_parts_if(mail, &block)
    removed_parts = []

    mail.parts.each do |part|
      if yield(part)
        mail.parts.delete(part)
        removed_parts << part
      end

      if part.mime_type == 'message/rfc822'
        sub_mail, sub_removed_parts = delete_parts_if(Mail.new(part.decoded), &block)
        part.body = sub_mail.encoded
        removed_parts += sub_removed_parts
      end
    end

    [mail, removed_parts]
  end

  def delete_attachment_parts(mail)
    delete_parts_if(mail, &:attachment?)
  end
end
```

TylerRick:
@jeremy I've rebased this PR. Let me know if there's anything else I can do to get it merged.

weilandia:
Hello, I'm hoping to use the `recursive_each` method to parse emails with `message/rfc822` parts, but I'm seeing it hasn't made it into a stable release.

Any suggestions for parsing those emails?  I'm thinking I'll just monkey-patch that method onto `PartsList` for now, but would be great to hear if that's not a good use case of the method.

PR #865, opened 2015-03-19
An invalid body encoding should still round trip

kjg:
I don't think the `if transfer_encoding == encoding` was adding anything
because the following logic would always error if dec was nil

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -6,6 +6,7 @@
 * #789 - Fix encoding collapsing not dealing with multiple encodings in 1 line (grosser)
 * #775 - avoid failed encodings / stop bad charsets early (grosser)
 * #849 - Handle calling Part#inline? when the Content-Disposition field couldn't be parsed (kjg)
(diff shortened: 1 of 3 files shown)


kjg:
Failure is due to a newer version of appraisal that doesn't support ruby 1.8.7

bf4:
Hmm.. we should lock that down, then

kjg:
@bf4, I agree. I also made a comment on the appraisal github. I'll follow up over there to see if I can get it fixed since they shouldn't be breaking backwards compatibility in a bugfix version bump.

That being said, this PR is tangential to all that. Please consider this outside the appraisal issue.

kjg:
This now passes against the latest master!

kjg:
I believe this is ready for merge. Is there anything else I can do to wrap this up? 

If you all are swamped and are interested in having some help, I'm willing and available to help out with PRs.

jeremy:
Thanks @kjg 

We'd love your help validating, reviewing, and giving feedback on PRs & issues, for sure! :heart:

PR #866, opened 2015-03-23
Support decoding message bodies with non-Ruby-standard charsets

jeremy:
We already have support for these additional charsets in message headers but didn't use the same code path for message bodies themselves.

Adds support for GB2312, ks_c_5601-1987, ISO-2022-JP-KDDI, ISO-8859-8-I, and more.


jeremy:
iconv doesn't natively recognize a bunch of these charsets, either. We'd need an alias mapping here like we use for for Ruby encodings.

PR #868, opened 2015-03-30
Use the Ruby19.charset_encoder when decoding message bodies

kjg:
To maintain consistency with param decoding


kjg:
How do you feel about this change? I think it makes for a nice simplification. It also allows for use of custom encoders (like the BestEffortCharsetEncoder) when figuring out the body's encoding. This provides consistency across headers and body encoding.

jeremy:
Yep yep! Check out #866 :grin:   Addresses failed body parsing despite having charsets we know about.

kjg:
That's awesome, I wish I had seen the original #866 before doing this.

Merged #868 <https://github.com/mikel/mail/pull/868>.

—
Reply to this email directly or view it on GitHub
<https://github.com/mikel/mail/pull/868#event-281595083>.

Issue #864, opened 2015-03-18
include BCC in encoded output based on configuration flag


grossadamm:
I'm trying to get the BCC included in the RFC 2822 format output for use with the gmail api. 

I completely understand the reasoning for not including the bcc by default in Issue #426, however, I still need it to be present so I can let gmail know as I'm not sending the individual emails out myself. What is the best way for me to go about doing so. 

I'm perfectly content code diving and adding the functionality that will add the bcc address to the encoded output only if some config flag is set. And I would of course create a corresponding pull request, but I don't want to reinvent the wheel if something similar already exists.

bf4:
You can monkey patch in the meantime to something like (untested code)

```ruby
class Mail::BccField
  def self.supress_encoding=(supress_encoding)
    @supress_encoding = supress_encoding
  end
  def self.supress_encoding
    @supress_encoding || self.supress_encoding = true
  end
  def encode
     if self.class.supress_encoding
       ''
     else
       do_encode(CAPITALIZED_FIELD)
    end
  end
end
```

and then `Mail::BccField.supress_encoding = false`

grossadamm:
Ok, so I'm guessing a way doesn't exist. I'll attempt to monkey-patch for now. Thanks!

bf4:
Would yo mind elaborating on where in your workflow with gmail the bug, with respect to gmail needing a bcc field, shows up?

grossadamm:
For the gmail api, to send an email you HTTP POST to the api an rfc2822 format email encoded in base64. [Api reference](https://developers.google.com/gmail/api/v1/reference/users/messages/send). The api supports the bcc field, however when I'm generating the rfc2822 format email via this gem the bcc field is not included. 

I understand the reason why the bcc field is being stripped out in the normal case, but in my arguably abnormal case, the stripping of the field makes it impossible for me to bcc messages using this gem to generate the rfc2822 format.

Exoth:
Yep, I need such an option to. Just an argument for to_s would be great for me.

grossadamm:
@bf4 I assume you are a busy person and accept you probably won't have the time to look at this let alone merge it for some time.

But whenever you get a chance, could you please review my pull request and see if there's anything you would change?

Thanks for the help!

bf4:
Cc @jeremy

jeremy:
[...]

Issue #426, opened 2012-08-13
BCC is ignored


oprogfrogo:
Rails 3.1.4
Ruby 1.9.3
mail (2.3.3)
actionmailer (3.1.4)

1.9.3p194 :008 > mail = Mail.new do
1.9.3p194 :009 >     from 'oprogfrogo@gmail.com'
1.9.3p194 :010?>   to 'oprogfrogo@gmail.com'
1.9.3p194 :011?>   bcc 'oprogfrogo@gmail.com'
1.9.3p194 :012?>   end
 => #<Mail::Message:70320662520600, Multipart: false, Headers: <From: oprogfrogo@gmail.com>, <To: oprogfrogo@gmail.com>, <Bcc: oprogfrogo@gmail.com>> 
1.9.3p194 :013 > mail.to_s
 => "Date: Mon, 13 Aug 2012 14:55:34 -0700\r\nFrom: oprogfrogo@gmail.com\r\nTo: oprogfrogo@gmail.com\r\nMessage-ID: 502977d6ab02d_17c6c3ff4cd434cdc4346a@Simon-Lees-iMac.local.mail\r\nMime-Version: 1.0\r\nContent-Type: text/plain\r\nContent-Transfer-Encoding: 7bit\r\n\r\n" 
1.9.3p194 :014 >

PR #872, opened 2015-04-15
Add setting suppress_encoding to BccField

grossadamm:
This pull request addresses issue #864.

Add suppress_encoding to BccField to enable encoding the bcc field. By default, the Bcc field is still not encoded for reasons listed in #426 and thus this does not break backwards compatibility.

Tested under:
1.9.3-p551
2.1.5
2.2.0
jruby-1.7.18

### DIFF
--- a/lib/mail/fields/bcc_field.rb
+++ b/lib/mail/fields/bcc_field.rb
@@ -43,9 +43,21 @@ module Mail
       self
     end
     
-    # Bcc field should never be :encoded
+    def self.suppress_encoding=(suppress_encoding)
+      @suppress_encoding = suppress_encoding
+    end
+
+    def self.suppress_encoding
+      defined?(@suppress_encoding) ? @suppress_encoding : self.suppress_encoding = true
+    end
+
+    # Bcc field should not be :encoded by default
     def encoded
-      ''
+      if self.class.suppress_encoding
+        ''
+      else
+        do_encode(CAPITALIZED_FIELD)
+      end
     end
     
     def decoded
--- a/lib/mail/fields/bcc_field.rb
+++ b/lib/mail/fields/bcc_field.rb
@@ -43,20 +43,20 @@ module Mail
       self
     end
     
-    def self.suppress_encoding=(suppress_encoding)
-      @suppress_encoding = suppress_encoding
+    def include_in_headers=(include_in_headers)
+      @include_in_headers = include_in_headers
     end
 
-    def self.suppress_encoding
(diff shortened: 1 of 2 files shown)


Exoth:
Is it supposed to be used globally by an initializer?
The problem is that I need two behaviors in one app (I'm sending emails both through sendmail and gmail). Changing the option right before and right after I send a gmail message looks strange to me. That's why an argument for `to_s` method looks as a better option for me.

jeremy:
Thanks @grossadamm. Made some comments @ https://github.com/mikel/mail/issues/864#issuecomment-93648760

The root issue is that we make the message BCC field do double duty as a real message header (which should be encoded when we set it) and as a list of envelope recipients (which aren't part of the message).

As a transitional step, we could mark the BCC field to indicate that we want to encode it.

grossadamm:
@Exoth, the new commit would allow you to configure per message if the bcc field is included in the headers. Each message would need to have the configuration value set as each new message would default to not including the bcc field. Would that work for you?

Exoth:
So is this gonna work like this: `message.bcc.include_in_headers = true`?
That looks perfect for me. Thanks.

grossadamm:
@Exoth correct

grossadamm:
@Exoth So apparently I fail at reading comprehension lately. `message.bcc` 
>Returns the Bcc value of the mail object as an array of strings of address specs.

So there is no way to access `include_in_headers` right on `message. I'm looking into what I need to do to make the setting more accessible.

This pull request is essentially unusable at the moment, my apologies to everyone while I get this straightened out.

grossadamm:
For a laugh, you can do the following and it will work

`message.header.fields.select {|f| f.name == 'Bcc'}.each { |f| f.include_in_headers = true }`

grossadamm:
Aha! After more reading of comments I learned that instead of `message.bcc.include_in_headers = true` you can do `message[:bcc].include_in_headers`. And that works as the pull request is currently written.

Would `message[:bcc].include_in_headers = true` that be acceptable to you @Exoth, @jeremy ?

Exoth:
Looks usable to me, though may be not the most obvious solution.

grossadamm:
Is there anything further I need to do here?

taavo:
This PR would be useful for me as well. My (eccentric) use case is that I have to generate RFC2822 email documents and post them to an API, and that API supports BCC. Rather than monkey patching, my current (shameful) implementation is to inject the BCC line manually.

Exoth:
Great. It would be nice if a new version of gem is released.

PR #874, opened 2015-04-19
Avoid violating RFC 2821 line length limits

### LINKED ISSUE #316: 7bit encoding line length
Hi,

For emails with a text/html part it appears that Mail defaults to using 7bit encoding for the text/html part.  When encoded is called on this part it does not break up the content into lines of 1000 characters.  It replaces "\n" or "\r" with "\r\n", but does not break the content into multiple lines.  RFC2045 implies that 7bit encoding still needs to represent data in 1000 character lines.

"7bit data" refers to data that is all represented as relatively short lines with 998 octets or less between CRLF line separation sequences [RFC-821].

Is there a way to make the 7bit encoded html part adhere to this 1000 character per line spec?  Alternatively, is there a way to force text/html parts to use quoted-printable encoding so that it is broken into multiple lines correctly?

Thanks,
Mark

pushrax:
Fixes #316.

RFC 2821 specifies that SMTP lines must not be longer than 1000 characters. In this case, quoted-printable or base64 should be used rather than 8bit or 7bit.

In Shopify we are running into an issue where user-written templates are generating long HTML lines, which causes Postfix to split them automatically at the 998-char boundary. Occasionally a URL or something important is split, leading to broken HTML rendering.

### DIFF
--- a/lib/mail/encodings/8bit.rb
+++ b/lib/mail/encodings/8bit.rb
@@ -26,7 +26,7 @@ module Mail
       end
 
       # Per RFC 2821 4.5.3.1, SMTP lines may not be longer than 1000 octets including the <CRLF>.
-      def self.should_transport?(str)
+      def self.compatible_input?(str)
         !str.lines.find { |line| line.length > 998 }
       end
 
--- a/lib/mail/encodings/8bit.rb
+++ b/lib/mail/encodings/8bit.rb
@@ -25,7 +25,7 @@ module Mail
         1.0
       end
 
-      # Per RFC 2921 4.5.3.1, SMTP lines may not be longer than 1000 octets including the <CRLF>.
+      # Per RFC 2821 4.5.3.1, SMTP lines may not be longer than 1000 octets including the <CRLF>.
       def self.should_transport?(str)
         !str.lines.find { |line| line.length > 998 }
       end
--- a/lib/mail/encodings/8bit.rb
+++ b/lib/mail/encodings/8bit.rb
@@ -25,6 +25,11 @@ module Mail
         1.0
       end
 
+      # Per RFC 2921 4.5.3.1, SMTP lines may not be longer than 1000 octets including the <CRLF>.
+      def self.should_transport?(str)
+        !str.lines.find { |line| line.length > 998 }
+      end
+
       Encodings.register(NAME, self) 
     end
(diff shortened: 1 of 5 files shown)


pushrax:
@mikel @jeremy ping for review.

lib/mail/encodings/8bit.rb
```diff
@@ -25,6 +25,11 @@ def self.cost(str)
         1.0
       end
 
+      # Per RFC 2821 4.5.3.1, SMTP lines may not be longer than 1000 octets including the <CRLF>.
+      def self.should_transport?(str)
```
BlakeMesdag: probably better named as `compatible_transport?` to indicate the purpose more clearly

lib/mail/encodings/transfer_encoding.rb
```diff
@@ -23,30 +23,39 @@ def self.cost(str)
...
+            self.can_transport?(enc) && enc.can_encode?(source_encoding)
+          end
+
+          best = nil
+          best_cost = nil
+
+          choices.each do |enc|
+            # If the current choice cannot be transported safely,
+            # give priority to other choices but allow it to be used as a fallback.
+            this_cost = enc.cost(str) if enc.should_transport?(str)
```
BlakeMesdag: `next unless enc.should_transport?` perhaps?

boourns:
:+1:

lib/mail/encodings/transfer_encoding.rb
BlakeMesdag: actually we can avoid this by verifying the `should_transport?` above as the last part of the conditional in `choices = …`

pushrax: That would work if we want to disable the encoding completely, but isn't the same as what the comment says (infinitely low priority). Maybe disabling would be better though?

BlakeMesdag: ah right, missed that part

wireframe:
any chance of this getting merged in?  I'm hitting this same issue and would love to see it finally resolved.

schiza:
@jeremy bump? This would be super useful merged ;)

sponomarev:
@pushrax Thanks a lot! So weird problem.

PR #882, opened 2015-05-27
Allow :filename option when using Mail::FileDelivery

kirikak2:


jeremy:
Good use for the `:logger` delivery method! #856

PR #891, opened 2015-07-30
Original-Envelope-Id support for delivery status reports

evserykh:


jeremy:
Thanks @evserykh! Let's leave this on the delivery status part rather than delegating its full interface.

Issue #889, opened 2015-07-23
smtp_envelope_from should allow NULL(<>) case.


kirikak2:
Mail library raises ArgumentError when smtp_envelope_from is nil in `check_delivery_params`.

I think smtp_envelope_from should allow NULL case. (ex. error mail)
What is the point of reject NULL case?

jeremy:
#896

PR #896, opened 2015-08-06
Feature/allow blank in smtp_envelope_from

kirikak2:
see #889

### DIFF
--- a/lib/mail/check_delivery_params.rb
+++ b/lib/mail/check_delivery_params.rb
@@ -1,10 +1,6 @@
 module Mail
   module CheckDeliveryParams
     def check_delivery_params(mail)
-      if mail.smtp_envelope_from.blank?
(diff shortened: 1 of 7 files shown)


jeremy:
+1 to allow blank SMTP `MAIL FROM`. Could you update the specs?

kirikak2:
Thank you for your response.
Ofcourse. I'll update specs soon.

kirikak2:
@jeremy It ready to merge.

jeremy:
Thanks @kirikak2. This flipped all the tests to say "doesn't raise an exception." Rather than testing that we don't raise an error, let's test the expected results. What happens when we send a message with no SMTP envelope From?

jeremy:
Rather than omit the envelope From address entirely, we can provide a null sender explicitly:
```ruby
smtp_envelope_from Mail::Constants::NULL_SENDER

# or
smtp_envelope_from '<>'
```

See #745

Issue #897, opened 2015-08-11
Mail::SendMail shellescape broken


benmmurphy:
https://github.com/mikel/mail/blob/master/lib/mail/network/delivery_methods/sendmail.rb#L85

So when invoking sendmail it takes the to address and escapes it and then puts it in double quotes. However, only $, `, ", and \ and newline are supported by the \ escape in bash in double quotes. So to addresses will be mangled and contain an extra \\.

For example:

    ~$ echo "foo\'bar@.com"
    foo\'bar@.com

From Bash manpage:

       Enclosing characters in double quotes preserves the literal value of all characters within the quotes, with the exception of $, `, \, and, when  history  expansion
       is  enabled,  !.  The characters $ and ` retain their special meaning within double quotes.  The backslash retains its special meaning only when followed by one of
       the following characters: $, `, ", \, or <newline>.  A double quote may be quoted within double quotes by preceding it  with  a  backslash.   If  enabled,  history
       expansion will be performed unless an !  appearing in double quotes is escaped using a backslash.  The backslash preceding the !  is not removed.

I'm not sure why we are using shellescape here and not execv style execution which would appear to be much safer.

jeremy:
#898

PR #898, opened 2015-08-12
add use_args option to SendMail delivery method

benmmurphy:
Fix for https://github.com/mikel/mail/issues/897

add use_args option to SendMail delivery method

if use_args is true then SendMail will invoke
the sendmail binary using execv style argument
passing instead of shell style argument passing.

this fixes issues where email addresses are
corrupted by shell escaping. this is an option
because it breaks compatibility when the
:arguments option is specified because the
:arguments option previously took a string of
arguments while when passing execv style we need
to take the option as an array of arguments.

this also breaks compatibility with the previous
behaviour because it now raises an error when
sendmail generates a non-zero exit code.

it also breaks compatibility with the previous
behaviour because use_args will not work with
older ruby versions. older ruby versions don't
support redirected stderr from popen and this is
probably why shell style argument passing was
originally used.

### DIFF
--- a/lib/mail/network/delivery_methods/sendmail.rb
+++ b/lib/mail/network/delivery_methods/sendmail.rb
@@ -40,9 +40,21 @@ module Mail
   class Sendmail
     include Mail::CheckDeliveryParams
 
+    class DeliveryError < StandardError
+    end
+
     def initialize(values)
+
+      arguments = "-i"
+
+      if values[:use_args]
+        arguments = ["-i"]
+      end
+
       self.settings = { :location       => '/usr/sbin/sendmail',
-                        :arguments      => '-i' }.merge(values)
+                        :arguments      => arguments,
+                        :use_args => false }.merge(values)
+
     end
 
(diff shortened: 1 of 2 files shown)


benmmurphy:
There is some really weird behaviour with travis and the specs I've added. 

So the tests I added where they weren't meant to raise DeliveryError were raising DeliveryError which presumably meant system("true") was returning a non-zero status. (https://travis-ci.org/mikel/mail/builds/75250035)

So I changed:

      if $?.exitstatus != 0
        raise DeliveryError
      end

to:

      if $?.exitstatus != 0
        raise DeliveryError.new("Bad Exit Status: #{$?.exitstatus}")
      end

to get a better idea of what was going on and then the builds started passing and this is not reproducible locally with similar ruby versions.

jeremy:
+1 for switching to execv style. It'd be interesting to see whether we could make this the default, even. Fall back to string style if `settings[:arguments]` is provided and can't be trivially converted to an argv array.

PR #901, opened 2015-08-16
Allow string assignment for text_part and html_part

taavo:
As a newcomer to this gem, when I tried to create a multipart message, my first attempt looked like this:

```
mail = Mail.new
mail.text_part = 'Some text'
mail.html_part = '<b>Some text</b>'
```

This, of course, doesn't work, because the two setters above are expecting to receive Mail::Message objects. I didn't like the two alternatives currently available, `mail.text_part = Mail::Part.new(body: 'Some text')` and `mail.text_part { body 'Some text' }`, because the first is too verbose and the second prohibits access to instance variables/methods from outside of Mail, so I added the two lines of code to make the above behave as expected.

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -1669,6 +1669,8 @@ module Mail
     def html_part=(msg)
       # Assign the html part and set multipart/alternative if there's a text part.
       if msg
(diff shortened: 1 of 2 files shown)


kjg:
:+1: This mirrors the block style syntax for setting text_part and html_part

PR #907, opened 2015-09-10
make Mail::ContentDispositionField work with nil value

kjg:
I've been running into an issue where a content disposition field fails to fully parse due to the filename being in an unspecified encoding. When this happens the ContentDispositionField ends up being crated with a nil value and then when .parameters is called on the field, it raises an error. Instead the parameters should just be empty.


kjg:
@jeremy @bf4 , would you mind looking at this please?

jeremy:
Thanks @kjg !

PR #915, opened 2015-10-05
Fix for nil content-type in 2.5

zeepeeare:
I've encountered some emails which have a content-type but it is nil, which causes a NoMethodError on line 48. Replacing this with the built-in method `to_s` will produce an empty string instead of an error, allowing this to pass silently. I have included a spec to demonstrate this behavior.

I understand this has been fixed in newer versions of this gem (2.6), but for other reasons we must continue using 2.5-stable and as such this is a PR to that branch.

### DIFF
--- a/lib/mail/parts_list.rb
+++ b/lib/mail/parts_list.rb
@@ -44,7 +44,7 @@ module Mail
   private
 
     def get_order_value(part, order)
-      if part.respond_to?(:content_type)
+      if part.respond_to?(:content_type) && !part[:content_type].nil?
         order.index(part[:content_type].string.downcase) || 10000
       else
         10000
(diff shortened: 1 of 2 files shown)


jeremy:
Looks like this change on master: https://github.com/mikel/mail/commit/60084517efb523759c70453855dc0d088be487a8

Could you backport that commit (`git cherry-pick 60084517efb523759c70453855dc0d088be487a8`) to `2-5-stable` instead of introducing a new code path?

zeepeeare:
@jeremy is there anything I should do for this build to pass?

jeremy:
@zeepeeare Fixed build https://github.com/mikel/mail/pull/1004. Please rebase.

(You can omit the previous commits instead of undoing them, too. Check out interactive rebase with `git rebase -i origin/2-5-stable`.)

zeepeeare:
Thanks for the suggestions!

PR #918, opened 2015-10-14
Account for possibility of absent delivery-status headers

kjg:


kjg:
@jeremy when you have a chance would you mind taking a look at this one?

jeremy:
Thanks @kjg!

Issue #902, opened 2015-08-17
Invalid utf-8 strings


sk-:
In many places invalid UTF-8 strings are returned. And in order to use them later (for example to encode them as JSON) you first need to `scrub` them.

How to reproduce:

``` ruby
Mail.new(subject: "=?windows-1258?Q?=F3?=").subject.valid_encoding?
```

james-lawrence:
+1 these shouldn't be returned as UTF8 unless they have been converted properly. Do not lie to the program about the encoding.
```ruby
    def Ruby19.b_value_decode(str)
      match = str.match(/\=\?(.+)?\?[Bb]\?(.*)\?\=/m)
      if match
        charset = match[1]
        str = Ruby19.decode_base64(match[2])
        str = charset_encoder.encode(str, charset)
      end
      decoded = str.encode(Encoding::UTF_8, :invalid => :replace, :replace => "")
      decoded.valid_encoding? ? decoded : decoded.encode(Encoding::UTF_16LE, :invalid => :replace, :replace => "").encode(Encoding::UTF_8)
    rescue Encoding::UndefinedConversionError, ArgumentError, Encoding::ConverterNotFoundError
      warn "Encoding conversion failed #{$!}"
      # should either re-raise, or don't force the encoding.
      str.dup.force_encoding(Encoding::UTF_8)
    end
```

```ruby
    def Ruby19.q_value_decode(str)
      match = str.match(/\=\?(.+)?\?[Qq]\?(.*)\?\=/m)
      if match
        charset = match[1]
        string = match[2].gsub(/_/, '=20')
        # Remove trailing = if it exists in a Q encoding
        string = string.sub(/\=$/, '')
        str = Encodings::QuotedPrintable.decode(string)
        str = charset_encoder.encode(str, charset)
        # We assume that binary strings hold utf-8 directly to work around
        # jruby/jruby#829 which subtly changes String#encode semantics.
        str.force_encoding(Encoding::UTF_8) if str.encoding == Encoding::ASCII_8BIT
      end
      decoded = str.encode(Encoding::UTF_8, :invalid => :replace, :replace => "")
      decoded.valid_encoding? ? decoded : decoded.encode(Encoding::UTF_16LE, :invalid => :replace, :replace => "").encode(Encoding::UTF_8)
    rescue Encoding::UndefinedConversionError, ArgumentError, Encoding::ConverterNotFoundError
      warn "Encoding conversion failed #{$!}"
      # should either re-raise, or don't force the encoding.
[...]

PR #921, opened 2015-10-20
do not lie about the strings encoding

james-lawrence:
one possible way to address https://github.com/mikel/mail/issues/902
this is more of pointing out the issue further than an actual proposed fix.

### DIFF
--- a/lib/mail/encodings.rb
+++ b/lib/mail/encodings.rb
@@ -316,17 +316,16 @@ module Mail
 
           # emit the substring before the beginMarker (if any)
           if beginMarker != 0
-            yielder << s[0..beginMarker-1]
+            yielder << s[0..beginMarker-1].rstrip
           end
 
           # emit the encoded characters section
           yielder << s[beginMarker..offset]
 
           # take the tail end of the string for further processing
-          s = s[offset..-1].strip
+          s = s[offset..-1]
         end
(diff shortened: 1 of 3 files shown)


jeremy:
Think this is a good idea. Pragmatic choice.

bf4:
:+1: Meanwhile, in terms of a fix, I covered all encoding cases in rspec-support

- https://github.com/rspec/rspec-support/blob/81e33d4ed3/lib/rspec/support/encoded_string.rb
- https://github.com/rspec/rspec-support/blob/81e33d4ed3/spec/rspec/support/encoded_string_spec.rb

Perhaps we/I should extract parts of that

james-lawrence:
have some further tests/fixes. this code is broken when emails contain array header fields with multiple encodings. the gsub fails and blows up wonderfully if a string is say utf-7 encoded.

james-lawrence:
further investigation leads me to say that the multiple encodings isn't a problem unless ruby can't correctly decode once of the encoding, at which point it should blow up.

bf4:
Extracted https://github.com/bf4/encoded_string with rspec's blessing.. It's well-tested, but I haven't done much with it on its own. Feel free to try it out

james-lawrence:
dead pr.

PR #923, opened 2015-11-03
Allow for combination of non-ascii and quotes

averell23:
See the spec for the original problem. Basically,
Encodings.encode_non_usascii(address, charset) got lost if the
address did include double quotes that were escaped like this:
`\"` and broke the address up incorrectly.

This seems to fix the immediate problem, although I guess there
may be more complications to this.


averell23:
To add a bit more background: This is derived of an actual example from our production system. It might not be the most elegant solution, and I didn't re-read RFC822 on how the quote-escaping should be handled. 

However a test mail did send correctly with the examples given.

PR #924, opened 2015-11-06
attachment matcher for mail

schepedw:
Allow testing for attachments

### DIFF
--- a/lib/mail/matchers/has_sent_mail.rb
+++ b/lib/mail/matchers/has_sent_mail.rb
@@ -42,15 +42,25 @@ module Mail
 
       def bcc(recipient_or_list)
         @blind_copy_recipients ||= []
-
-        if recipient_or_list.kind_of?(Array)
-          @blind_copy_recipients += recipient_or_list
-        else
-          @blind_copy_recipients << recipient_or_list
-        end
+        @blind_copy_recipients.concat(Array(recipient_or_list))
         self
       end
 
+      def with_attachments(attachments)
+        @attachments ||= []
+        @attachments.concat(Array(attachments))
+        self
+      end
+
+      def with_no_attachments
+        @having_attachments = false
+        self
+      end
+
+      def with_any_attachments
+        @having_attachments = true
+        self
+      end
 
       def with_subject(subject)
         @subject = subject
@@ -95,8 +105,10 @@ module Mail
 
       def filter_matched_deliveries(deliveries)
         candidate_deliveries = deliveries
-
-        %w(sender recipients copy_recipients blind_copy_recipients subject subject_matcher body body_matcher).each do |modifier_name|
+        modifiers =
+          %w(sender recipients copy_recipients blind_copy_recipients subject
+          subject_matcher body body_matcher having_attachments attachments)
+        modifiers.each do |modifier_name|
           next unless instance_variable_defined?("@#{modifier_name}")
           candidate_deliveries = candidate_deliveries.select{|matching_delivery| self.send("matches_on_#{modifier_name}?", matching_delivery)}
         end
@@ -126,6 +138,17 @@ module Mail
 
       def matches_on_subject_matcher?(delivery)
         @subject_matcher.match delivery.subject
+      end
+
+      def matches_on_having_attachments?(delivery)
+        @having_attachments && delivery.attachments.any? ||
+          (!@having_attachments && delivery.attachments.none?)
+      end
+
+      def matches_on_attachments?(delivery)
+        @attachments.each_with_index.inject( true ) do |sent_attachments, (attachment, index)|
+          sent_attachments &&= (attachment === delivery.attachments[index])
+        end
       end
 
       def matches_on_body?(delivery)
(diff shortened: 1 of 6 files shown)


schepedw:
@mikel can you review / merge this?

spec/matchers_spec.rb
```diff
@@ -3,15 +3,19 @@
...
-    Mail.deliver do
-      from    'phil@example.com'
-      to      ['bob@example.com', 'fred@example.com']
-      cc      ['dad@example.com', 'mom@example.com']
-      bcc     ['alice@example.com', 'sue@example.com']
-      subject 'The facts you requested'
-      body    'Here are the facts you requested. One-onethousand, two-onethousand.'
-    end
+  def test_mail
+    return @mail unless @mail.nil?
```
jeremy: Can use `let(:test_mail) { … }` rather than doing early-return memoization here

lib/mail/matchers/has_sent_mail.rb
```diff
@@ -128,6 +134,12 @@ def matches_on_subject_matcher?(delivery)
         @subject_matcher.match delivery.subject
       end
 
+      def matches_on_attachments?(delivery)
+        return false if delivery.attachments.empty?
+        return true if @attachments.empty?
+        @attachments.all? { |attachment| delivery.attachments.include?(attachment) }
+      end
+
```
jeremy: Can trade the early returns for a boolean expression:
```ruby
delivery.attachments.any? && @attachments.all? { |a| delivery.attachments.include? a }
```
(When `@attachments` is `[]`, `#all?` will always be `true` so we needn't check that separately.)

spec/matchers_spec.rb
```diff
@@ -137,6 +141,22 @@ def send_test_email
...
+
+    context 'and non-matching attachments' do
+      it { is_expected.not_to have_sent_email.with_attachments('no_match') }
+    end
+
+    context 'and any attachments' do
+      it { is_expected.to have_sent_email.with_attachments }
+    end
+  end
+
```
jeremy: Need a test for multiple attachment arguments, too.

lib/mail/matchers/has_sent_mail.rb
```diff
@@ -51,6 +51,10 @@ def bcc(recipient_or_list)
         self
       end
 
+      def with_attachments(*attachments)
+        @attachments = attachments
+        self
+      end
```
jeremy: 1. This API naming suggests that you should pass an array of attachments as the argument, which is how the other methods behave.
2. Chaining multiple `with_attachments` calls will overwrite the previous calls rather than accumulating them.

Consider something like
```ruby
def with_attachments(attachment_or_list)
  @attachments ||= []
  @attachments.concat Array(attachment_or_list)
end
```

+test for chained `with_attachments` calls.

schepedw: passing `attachment_or_list` instead of `*attachments` disallows us to pass no argument to this matcher.  I like the splat better, what are your thoughts?

jeremy: There's more than one route to implement this in Ruby. You can use splats so long as this method behaves like the rest of the API.

Is the idiomatic way to pass an "any" matcher rather than treating empty splat args as an implied "any," e.g.
```ruby
is_expected.to have_sent_email.with_attachments(any_attachment)
```

schepedw: Ah, I didn't previously understand that you were pointing towards [matching arguments](https://relishapp.com/rspec/rspec-mocks/v/3-3/docs/setting-constraints/matching-arguments). If I were to impement `any_attachment`, or `an_attachment_with_filename(filename)`, what do you envision chaining them them together would look like? or would there never be intent to create a test like
` it { is_expected_to have_sent_email.with_attachments(any_attachment).with_attachments(an_attachment_with_filename('foo')) }`

schepedw:
@jeremy These failing tests are weird. They happen for me locally on a fresh checkout of master:
```
$ git remote -v
origin	https://github.com/mikel/mail.git (fetch)
origin	https://github.com/mikel/mail.git (push)
schepedw	https://github.com/schepedw/mail.git (fetch)
schepedw	https://github.com/schepedw/mail.git (push)
$ git pull origin master 
From https://github.com/mikel/mail
 * branch            master     -> FETCH_HEAD
Already up-to-date.
$ chruby
 * rbx-2.5.5
   ruby-2.2.2
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.

nothing to commit, working directory clean
$ bundle && bundle exec rspec
#bundle goes fine...
#a whole bunch of passing tests, then 
no such file to load -- ripper (LoadError)

Backtrace:
                 Rubinius::CodeLoader#load_error at kernel/common/code_loader.rb:445

#same stack trace as above travis failures.
```
Any insight? Am I missing something here?

schepedw:
Unrelated, can you review the latest implementation?

schepedw:
@jeremy tests are back to green

Thanks to @kjg for pointing me to https://github.com/rspec/rspec-support/pull/251

jeremy:
Nice work @schepedw!

PR #925, opened 2015-11-12
Fix non-us matching.

jroith:
The parser currently does not correctly work with non-us ascii. In particular, if a list contains multiple quoted addresses and the first one is us ascii, but a later one is not, the regular expression will not work properly.

For Example: `Mail::ToField.new('"Just Ascii" <just@ascii.com>, "Foo Bär" <x@example.com>').addrs`

This is broken upstream:
 `=> [#<Mail::Address:70340928511720 Address: |"Just Ascii\" <just@ascii.com>, \"Foo Bär" <x@example.com>| >]`

With the patch:

 `=> [#<Mail::Address:70340922281680 Address: |Just Ascii <just@ascii.com>| >, #<Mail::Address:70340922281660 Address: |"Foo Bär" <x@example.com>| >]`

### DIFF
--- a/lib/mail/encodings.rb
+++ b/lib/mail/encodings.rb
@@ -179,7 +179,7 @@ module Mail
       return address if address.ascii_only? or charset.nil?
       us_ascii = %Q{\x00-\x7f}
       # Encode any non usascii strings embedded inside of quotes
-      address = address.gsub(/(".*?[^#{us_ascii}].*?")/) { |s| Encodings.b_value_encode(unquote(s), charset) }
+      address = address.gsub(/("[^"]*?[^#{us_ascii}][^"]*?")/) { |s| Encodings.b_value_encode(unquote(s), charset) }
       # Then loop through all remaining items and encode as needed
       tokens = address.split(/\s/)
       map_with_index(tokens) do |word, i|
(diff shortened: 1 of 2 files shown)


lib/mail/encodings.rb
```diff
@@ -179,7 +179,9 @@ def Encodings.encode_non_usascii(address, charset)
       return address if address.ascii_only? or charset.nil?
       us_ascii = %Q{\x00-\x7f}
       # Encode any non usascii strings embedded inside of quotes
-      address = address.gsub(/(".*?[^#{us_ascii}].*?")/) { |s| Encodings.b_value_encode(unquote(s), charset) }
+      address = address.gsub(/(".*?.*?")/) do |q|
```
kjg: Instead of nested gsubs, what about `("[^"]*?[^#{us_ascii}][^"]*?")`

kjg:
Please see my comment about an alternative to nested gsubs. Please also add a test to `spec/mail/encodings_spec.rb` in the encode_non_usascii section

kjg:
an spec that might work 

```ruby
    it "should encode multiple combo of ascii then non-ascii correctly" do
      raw     = '"Mikel Lindsaar" <mikel@test.lindsaar.net>, "あdあ" <ada@test.lindsaar.net>'
      encoded = '"Mikel Lindsaar" <mikel@test.lindsaar.net>, =?UTF-8?B?44GCZOOBgg==?= <ada@test.lindsaar.net>'
      expect(Mail::Encodings.encode_non_usascii(raw, 'utf-8')).to eq encoded
    end
```

kjg:
@jeremy mind taking a look at this one when you get a chance? I think all the rbx failures are not related to the change, but I don't have the ability to re-run master to double check.

jeremy:
Fixed by #1074

PR #928, opened 2015-11-22
mail works with mime-types 3.0

halostatue:
- mime-types 3.0 was released today and the unit for mail pass with mime-types
  version 3. This is expected as there were only deprecated methods removed.
  This will enable other applications to upgrade.


bf4:
Shakes fist RUBINIUS!!!!. @halostatue Hi, nice meeting you last week :)

bf4:
Would you mind adding a `gemfile` for 3.0 for the Travis appraisals (and updating the .travis.yml to include it)? Thanks e.g. https://github.com/mikel/mail/blob/master/gemfiles/mime_types_2.latest.gemfile

halostatue:
I’ve done a few more than just the 3.0 appraisal. I added your missing versions from 2.4…2.6 including a variation for columnar loading, plus I added 2.99 (the last 2.x version for those stuck on 1.9.x). No clue if Travis is going to like this one or not, but it’s much more comprehensive than it used to be.

halostatue:
This is going to fail the build, but I’m not sure why the builds that have failed are failing (“appraisal ~> 1.0 not available for this machine”, for example).

halostatue:
OK. Some of these appear to be intermittent rubygems.org failures; what is clear is that we need to also include jruby-9.0.0.0 as part of this because it’s the only jruby that includes Ruby 2.0 support. This will, however, make your matrix go even wider.

bf4:
I reran the false negatives. Looks like the gemfiles/*.gemfile need to have `ruby-debug` removed?

deivid-rodriguez:
For some reason the install phase doesn't seem to be run in some cases. Maybe you can try to be more explicit:

```diff
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,6 @@ gemfile:
 
 sudo: false
 
-bundler_args: --path vendor/bundle
-
 script: bundle exec rake
 
 env:
@@ -38,6 +36,9 @@ env:
 before_install:
   - bundle config --local without local_development
 
+install:
+  - bundle install --path vendor/bundle
+
 matrix:
   exclude:
     - rvm: 1.8.7
```

bf4:
@deivid-rodriguez @halostatue I'm re-running CI.  Let's see what happens

deivid-rodriguez:
Oh, I see the error now. The gemfile specified in the failing cases, `gemfiles/mime_types_2.99.latest.gemfile`, does not exist, so the install phase can't run. @halostatue can probably fix that very easily.

deivid-rodriguez:
It's probably a typo and he meant `gemfiles/mime_types_2.latest.gemfile` instead of `gemfiles/mime_types_2.99.latest.gemfile`

halostatue:
I actually meant both. Modified and pushed.

halostatue:
And pushed again. This should cause it to be fixed. One note—there is nothing in your matrix to test JRuby 9k.

deivid-rodriguez:
You're almost there, only one failure to go. :)

.travis.yml
```diff
@@ -43,8 +50,28 @@ matrix:
...
+    - rvm: 1.8.7
+      gemfile: gemfiles/mime_types_2.99.gemfile
+    - rvm: 1.8.7
       gemfile: gemfiles/mime_types_2.latest.gemfile
     - rvm: 1.8.7
+      gemfile: gemfiles/mime_types_3.0.gemfile
+    - rvm: 1.9.2
+      gemfile: gemfiles/mime_types_3.0.gemfile
+    - rvm: 1.9.3
+      gemfile: gemfiles/mime_types_3.0.gemfile
```
bf4: Looks like jruby also needs to be excluded from the >= 3.0

>  Gem::InstallError: mime-types-data requires Ruby version >= 2.0.
An error occurred while installing mime-types-data (3.2015.1120), and Bundler
cannot continue.
Make sure that `gem install mime-types-data -v '3.2015.1120'` succeeds before
bundling.

halostatue: Will do. Should I add a jruby 9k set to the matrix?

bf4: Sure, why not :)

halostatue: OK. I’ll add that in the next day or so and we can get this merged finally. ;)

bf4: Maybe this? 

```yaml
matrix:
  include:
  - rvm: jruby-9.0.3.0
```

cargo culted from https://github.com/bundler/gemstash/pull/47/files#diff-354f30a63fb0907d4ad57269548329e3R5

though maybe set the JRUBY_OPTS in travis as

```yaml
  - rvm: jruby-9.0.3.0
    env: JRUBY_OPTS="$JRUBY_OPTS -J-Xmx1536m -J-XX:MaxPermSize=192m"
```

or as in  https://github.com/rspec/rspec-dev/pull/115

```yaml
  - rvm: jruby-9.0.3.0
    env: JRUBY_OPTS='--server -Xcompile.invokedynamic=false'
```
```yaml
    - rvm: jruby-19mode
      env: JRUBY_OPTS="$JRUBY_OPTS --debug"  # for simplecov
    - rvm: jruby-20mode
      env: JRUBY_OPTS="$JRUBY_OPTS --debug"  # for simplecov
    - rvm: jruby-head
      env: JRUBY_OPTS="$JRUBY_OPTS --debug"  # for simplecov
```

like in https://github.com/dkubb/sql/blob/36bc3d3834ed36e500fa8184c6c9f57c6f8af6b0/.travis.yml

ref: https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

`JRUBY_OPTS="--server -Dcext.enabled=false -Xcompile.invokedynamic=false"`

halostatue: Probably. I haven’t yet had a chance to do anything with this—next week is a partial hack week for us, so I think I’ll be able to finish some of this stuff then.

PR #931, opened 2015-11-23
Add capability to register and unregister failure handlers

j-concepcion:
Failure handlers are triggered when the email fails to send and an exception is raised.

Tested against:

- 1.8.7-p375, "mime-types-1.16" appraisal (newer specified versions of mime-types seem to require >= ruby 1.9.3)
- 1.9.3-p551 (tested all appraisals specified)
- 2.2.2 (tested all appraisals specified)


ramontayag:
I worked on this with @j-concepcion. Is there anything we can do to get this merged? Do you want us to test anything else?

ramontayag:
Just updated this to 1) get it mergeable again, 2) follow the format of changelog

jeremy:
Cool idea 👍

ramontayag:
Thanks! It looks like it's failing for some other reason.

PR #934, opened 2015-11-24
Don't let ActiveSupport exist after loading unicode tables.

### LINKED ISSUE #933: Mail with no ActiveSupport
Hi,

I'm having a problem with Mail loaded in an application with no ActiveSupport. Mail defines `ActiveSupport` ([here](https://github.com/mikel/mail/blob/master/lib/mail/multibyte/unicode.rb#L394-L400)), which gets us in to trouble when other gems looks for `ActiveSupports` and finds it and thinks that `ActiveSupport.on_load` is defined too.

To give a concrete example, here is an error I do get when I pull in Mail and Awesome print together with no ActiveSupport.

`awesome_print-1.6.1/lib/awesome_print.rb:29:in `<top (required)>': undefined method `on_load' for ActiveSupport:Module (NoMethodError)`

Can the definition of ActiveSupport be removed? I did see that it has something to do with the [mail/values/unicode_tables.dat](https://github.com/mikel/mail/blob/master/lib/mail/values/unicode_tables.dat). I tried to solve it myself and make a pull request, but I didn't come to any good working solution.

thhermansen:
This hopefully fixes #933, but it feels rather hackish. I guess the old code was dodgy as well, so.. 

Maybe someone comes up with a better solution.

### DIFF
--- a/lib/mail/multibyte/active_support_shim.rb
+++ b/lib/mail/multibyte/active_support_shim.rb
@@ -1,0 +1,7 @@
+unless defined?(ActiveSupport)
+  module ActiveSupport
+    module ShimInsertedByMail; end
+
+    Multibyte = Mail::Multibyte
+  end
+end
--- a/lib/mail/multibyte/unicode.rb
+++ b/lib/mail/multibyte/unicode.rb
@@ -339,6 +339,12 @@ module Mail
 
         # Loads the Unicode database and returns all the internal objects of UnicodeDatabase.
         def load
+          # The Marshal.load of unicode_tables.dat expects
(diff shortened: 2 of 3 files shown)


bf4:
So, this was addressed in an unmerged PR https://github.com/mikel/mail/pull/830

thhermansen:
Seems like #830 has the correct fix for this as well, as I didn't figure out how the database marshal dump was generated. I see that #830 includes generation, so I do like that to be merged in instead of this one.

To bad I didn't find it earlier before @jeremy asked me to give it a shot :) No worries.

thhermansen:
Btw: Thank you @bf4

Issue #879, opened 2015-05-22
matching_body matcher fails to check multipart emails


stefanosc:
Hello and thank you for the great work!
I am using the matchers in a rails project and I love them :smile: 

I found that `matching_body` only works on [mail.body.raw_source](https://github.com/mikel/mail/blob/master/lib/mail/matchers/has_sent_mail.rb#L136), if the mail is multipart (text/plain, text/html) as it is common these days then this matcher fails to check the body.

I thought it could be useful to either have an additional *arg(s) to pass to the method something like:
```ruby
have_sent_email.matching_body(/my regex/, :html, :text)
```

or have 2 additional matchers for `mail.html_part.body` and `mail.text_part.body`

I could take a crack at putting PR together if you think it would be useful :smile: 
Let me know.

Thanks again :heart:

nicolas-fricke:
I also had similar issues and once already monkey patched a method like this for my local project – I'd be happy to submit it as a PR :smiley:

jeremy:
Moved to #935

PR #935, opened 2015-11-24
Add multipart email support for HasSentEmailMatcher#matching_body

nicolas-fricke:
This can be used to test if the body of a multipart email contains the correct text in its individual body parts.

A similar behavior was proposed by @stefanosc in issue https://github.com/mikel/mail/issues/879.

Usage:

``` Ruby
# To check if text matches on both parts, text and HTML, add `:on_both_parts` as second argument to `#matching_body`
it { is_expected.not_to have_sent_email.matching_body(/text|HTML/i, :on_both_parts) }

# To check if either of the two parts match use `:on_either_part` as second argument:
it { is_expected.to have_sent_email.matching_body(/plain/i, :on_either_part) }

# To only check if a match is on the text part of a multipart email use `:on_text_part`:
it { is_expected.to have_sent_email.matching_body(/plain|non-existing/i, :on_text_part) }

# And to only check if a match is on the HTML part use `:on_html_part`:
it { is_expected.to have_sent_email.matching_body(/h1|body/i, :on_html_part) }
```

This has been successfully tested with the following rubies:

```
   jruby-1.7.19 [ x86_64 ]
   ruby-1.8.7-head [ i686 ]
   ruby-1.9.3-p551 [ x86_64 ]
   ruby-2.1.5 [ x86_64 ]
   ruby-2.2.1 [ x86_64 ]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -631,6 +631,28 @@ describe "sending an email" do
 
   it { should have_sent_email.matching_subject(/test(ing)?/) }
   it { should have_sent_email.matching_body(/h(a|e)llo/) }
+  
+  # Or if you use multipart emails...
+  
+  # ...can check if text matches on both parts, text and HTML
+  
+  it { is_expected.not_to have_sent_email.matching_body(/text|HTML/i) }
+  
+  # or same, but more explicit
+  
+  it { is_expected.not_to have_sent_email.matching_body(/text|HTML/i, :on_both_parts) }
+  
+  # ...can check if either of the two parts match by using `:on_either_part`
+   
+  it { is_expected.to have_sent_email.matching_body(/plain/i, :on_either_part) }
+  
+  # ...can only check if a match is on the text part of a multipart email by using `:on_text_part`
+  
+  it { is_expected.to have_sent_email.matching_body(/plain|non-existing/i, :on_text_part) }
+  
+  # ...and can only check if a match is on the HTML part by using `:on_html_part`
+  
+  it { is_expected.to have_sent_email.matching_body(/h1|body/i, :on_html_part) }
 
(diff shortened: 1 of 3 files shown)


jeremy:
Dig the idea, but the `multipart_match_on` arg feels idiosyncratic. Perhaps it should be part of the method call to create the matcher:
```ruby
have_sent_email.matching_text_part(/foo/).matching_html_part(/h1/).matching_multipart(/in both/)
```
I'm grasping for a case where we'd want to match against one part or the other. We'll either be looking for the same content in both parts or getting particular about what's in one of the parts.

nicolas-fricke:
I first had it as a separate method (though only for all multipart matchers and still using a second argument to specify the part). But when I wrote my first tests for `mail`, I had the same problem  @stefanosc mentioned in issue #879 – I wondered why `.matching_body` did not work.

But it is true that the `either` matcher probably won't really be all that useful. As a user, I probably either want to ensure that some text is in both parts or only in one specific, but then I care about in which it is. Therefore, if no one objects, I'll remove it.

Okay, so maybe we could have the more specific matchers and then in `.matching_body` check if it is a multipart message and then call the _both_ matcher (or maybe the _either_ matcher in this case?). Not entirely sure what the best approach might be here. Opinions?

jeremy:
> As for making .matching_body behave more like one may assume it does, we could maybe check wether the message is multipart and in that case call the both matcher

Agreed.

jeremy:
Check out with_html and with_text matchers from #664.

trafium:
@jeremy 
Specified PR does not address the possibility of partial matching on html part. Would another PR with this functionality be considered?

Issue #936, opened 2015-11-27
Remove monkey-patching?


jodosha:
Hello there, thank you for this great, and I mean **great** gem. :100: 

I'm wondering if there is a chance that you would accept a PR that removes monkey patching.

This technique is dangerous and can conflict with other gems that try to redefine the same set of methods. If you [search on GitHub](https://github.com/search?l=ruby&q=%22def+blank%22+language%3ARuby&ref=advsearch&type=Code&utf8=%E2%9C%93) for `blank?` definition, you'll get ~41,000 results. These gems doesn't even care to check if that method was previously defined.

This exposes `mail` to unforeseen compatibility problems with the Ruby ecosystem.

What are your thoughts about this? I can take care of this change, by submitting a PR.

Thanks in advance for your time and thanks for evaluating this request.

darrencauthon:
Wouldn't any library that defined ```blank?``` do so with the same basic concept of "blank-ness?"  

It seems that defining ```blank?``` on a class would actually make it **more** compatible with the Ruby ecosystem, as the concept of "blankness" as a true/false is a common idea.  Monkey-patching it into Object just sets that idea at a very low level, allowing other classes to implement it as they need.

mail's implementation looks to be the same as activesupport, which has become the standard for this implementation in the Ruby ecosystem (not saying whether that's good or bad).

jodosha:
@darrencauthon Hello and thanks for the quick reply :smile: 

> Wouldn't any library that defined blank? do so with the same basic concept of "blank-ness?"

Not really, look at [this](https://github.com/makevoid/bitcoin_exchange/blob/a39fb5b750cd8e2efef77c6bb10d9ef6d123484b/lib/monkeypatches/string.rb#L3) implementation from https://github.com/makevoid/bitcoin_exchange

```ruby
class String
  def blank?
    self == ""
  end
end
```

With the [one](https://github.com/mikel/mail/blob/master/lib/mail/core_extensions/string.rb#L25) from `mail`:

```ruby
class String
  def blank?
    self !~ /\S/
  end
end
```

And the [one](https://github.com/rails/rails/blob/7f18ea14c893cb5c9f04d4fda9661126758332b5/activesupport/lib/active_support/core_ext/object/blank.rb#L115) from ActiveSupport:

```ruby
class String
  BLANK_RE = /\A[[:space:]]*\z/

  def blank?
    BLANK_RE === self
  end
end
```

[...]

PR #937, opened 2015-11-30
Remove monkey patched blank

darrencauthon:
The start of a strategy to remove the monkey-patched ```blank?``` method.  This approach could be used as an example to remove other monkey patches / "core extensions."

This approach works, it's just time for me to get ready to go to work.  The tests that aren't passing yet are due to the tests tie to ```be_blank```.  Just have to thread this new helper through those as well, then test it.

One thing that surprised me was how nice it was to have the monkey patched ```blank?``` behavior consolidated into one method -- versus three separate monkey-patches.


darrencauthon:
This was meant to be an example for https://github.com/mikel/mail/issues/936 ... I'll close this for now unless this is actually an idea that the gem maintainers want to move forward with.

jeremy:
Generally positive on keeping our utility methods in-house. Reopening.
* Blankness isn't String-specific. Could live on `Mail::Utilities.blank?`
* Many of these cases may be more precise, since they're really `nil?` or `empty?` checks.

darrencauthon:
Ok, I'll do that.

* Yeah -- I noticed that it's not string specific as I was doing it, but I didn't want to stop.  I'll rename that module before it's ready.

* I was thinking the same thing, but I think it would be safer to consider this a refactor **only** and only change the monkey-patches to utility methods.  Logic changes can be incorporated easier after the refactor is done.

darrencauthon:
@jeremy @jodosha So I think I have this ```blank?``` method removed as a monkey patch.

These were mostly brain-less commits.  I swapped the ```blank?``` method with a module method, and then I swept through the project replacing the method calls.  So now the project works (the tests all pass) but String has one less method.

1__darrencauthon_darrens-macbook-air____darren_mail__zsh_
 
So where do we go from here?  Continue on with the rest of the extensions in this PR? Merge this and make the rest their own PRs?  Go through the logic, like @jeremy pointed out, and figure out if ```blank?``` is actually the right method to use?  Or just close? :smile:

darrencauthon:
Oh, and I must point out one thing:

**These could be considered breaking changes.**

Perhaps someone has Ruby code that depends on these core extensions?

The easiest solution is to tell them to bring in active support.  But these could, technically, be considered breaking changes.  Especially because the README mentions that this gem will create those methods.

jodosha:
@darrencauthon @jeremy Thanks for taking time to prepare this PR. :sparkles: 

I think this is a step forward, but module helper methods aren't handy if you want to remove the other monkey-patches from `String`. Think of `#to_crlf`, there are the same dangers of keeping it around that `#blank?` has. On GH there are [~2500 search results](https://github.com/search?utf8=%E2%9C%93&q=%22def+to_crlf%22+language%3ARuby&type=Code&ref=searchresults) for this method.

What do you think about `Mail::String` as an objects that wraps a string and acts like a string? See [`Lotus::Utils::String`](https://github.com/lotus/utils/blob/master/lib/lotus/utils/string.rb) and its [tests](https://github.com/lotus/utils/blob/master/test/string_test.rb) as an example.

> These could be considered breaking changes.
> 
> Perhaps someone has Ruby code that depends on these core extensions?

What's their API definition? Public or Private? :+1: To document the change in README anyway. :smile:

darrencauthon:
@jodosha @jeremy 

I don't like the idea of a String class that isn't a string.  A string is a string.  They look like this:  ```""```.  I don't want any code that looks like a String that's not a string.  And I don't want to have to call ```to_s``` on a "string" to get the string that I'm expecting from a... string.

I think @jodosha brought for a legitimate problem:  Monkey-patching can cause a problem.  This pull request offers what I believe is the **simplest** solution for the problem.  Make a method that behaves just like ```blank?```, then copy over the methods.  It's a brainless operation, just gotta turn the crank.

Putting in a new ```Mail::String``` class just complicates the matter, in my mind.  And I think it requires the same brainless operations, turning the crank... **if you are being safe**.  I hope nobody's considering devising a "cute" solution to this problem, internally using a new string class.  :fearful: 

But that's just my preference. :smile:  I'll continue to plod away and remove the rest of the core extensions in this fashion, but only if it's the path the gem maintainers want to go.

lib/mail/utilities.rb
```diff
@@ -221,5 +221,24 @@ def map_with_index( enum, &block )
...
 
+    # Returns true if the object is considered blank.
+    # A blank includes things like '', '   ', nil,
+    # and arrays and hashes that have nothing in them.
+    # 
+    # This logic is mostly shared with ActiveSupport's blank?
+    def self.blank?(string)
+      if string.is_a?(NilClass)
+        true
+      elsif string.is_a?(String)
```
jeremy: Will want to use `kind_of?` rather than `is_a?` on these to catch subclasses. Or use a `case` statement which'll use `#===` comparison.

jeremy:
This is great @darrencauthon. Thank you for pursuing it!

> Continue on with the rest of the extensions in this PR? Merge this and make the rest their own PRs? Go through the logic, like @jeremy pointed out, and figure out if blank? is actually the right method to use?

Merge this and make the other extensions their own PRs. Ditto for dialing back `blank?` in favor of `nil?` and `empty?`.

darrencauthon:
Sounds good

lib/mail/utilities.rb
darrencauthon: @jeremy I switched to ```kind_of?```, but I don't think it mattered... I think ```is_a?``` is a synonym.

http://stackoverflow.com/questions/3893278/ruby-kind-of-vs-instance-of-vs-is-a

darrencauthon:
So long as the CI passes, I think this is done.  I'll be moving on to the other bits.

jeremy:
Nice! Please rebase+squash to a single atomic commit, and we're good to merge.

darrencauthon:
The build is passing.  One test run failed, but it looks like a random glitch in one environment.  

https://travis-ci.org/mikel/mail/jobs/95953783

lib/mail/fields/common/common_date.rb
```diff
@@ -11,7 +11,7 @@ def default
     end
     
     def parse(val = value)
-      unless val.blank?
+      unless Utilities.blank?(val)
         @element = Mail::DateTimeElement.new(val)
```
bf4: ew, an unless else :)

jodosha:
@darrencauthon @jeremy Thanks for merging this and sorry for my late reply. I really appreciate your effort. :sparkling_heart: 

@darrencauthon:
> And I don't want to have to call to_s on a "string" to get the string that I'm expecting from a... string.

As long `Mail::String` respond to all the methods that `::String` respond, there is no need to invoke `#to_s`. There are rare cases where you need to explicitly do that: at the boundary between `mail` and other frameworks, only when the code that uses `mail` expects a `::String`.

Issue #944, opened 2015-12-18
Need patch for OSVDB-131677 against 2.5.x


mattolson:
A security advisory was issued for the mail gem recently: https://github.com/rubysec/ruby-advisory-db/blob/master/gems/mail/OSVDB-131677.yml

It indicates that the vulnerability was fixed in 2.6.0. However, actionmailer 3.2 (part of rails 3.2) has a dependency on `~> 2.5.4`. See https://github.com/rails/rails/blob/3-2-stable/actionmailer/actionmailer.gemspec#L23

According to https://github.com/rails/rails/pull/22631, the rails project is unwilling to bump the version to 2.6.0. 

How hard would it be to backport that fix to 2.5 and cut a new release?

cswilliams:
+1, would also be great to have a 2.4.x backport as well for rails 3.1.

wktk:
Here's a monkey patch.  You can also apply this when you can't update the gem version for some reasons.

```ruby
require 'mail'

if Gem::Version.new(Mail::VERSION.version) < Gem::Version.new('2.6')
  Mail::Field.class_exec do
    original_method = instance_method(:create_field)
    define_method(:create_field) do |name, value, charset|
      value = value.gsub(/[\r\n \t]+/m, ' ') if value.is_a?(String)
      original_method.bind(self).call(name, value, charset)
    end
  end
end
```

Davidslv:
Is there anything preventing this from being fixed and released?

sgerrand:
:wave: @jeremy: Are there any blockers to #949 and #950 being merged into their relevant version branch targets? I'm interested in getting this applied to the `2.5-stable` branch.

jeremy:
This is backporting a change that's not meant to fix the underlying vuln (with input validation on net/smtp) but happens to mask it with a coincidental behavior change. It was called out the sec advisory only because it bisected as the fix.

We can introduce stronger input validation at the mail handling layer and backport it without breaking compat, but 1. this isn't that change and 2. it won't conclusively address the original vuln, SMTP injection via email address.

Will keep this issue open to track resolution.

GUI:
@jeremy: Could you elaborate on what other input validations would be needed to address the vulnerability in older versions of the mail gem? I'd be happy to try and lend a hand getting things back-ported, since I'd also find this very useful in dealing with older Rails version

[...]

PR #950, opened 2016-01-07
Security fix for #944 for 2.5.x: Escape brake code in fields

wktk:
As described in #944, there is a security problem in parsing headers.
I backported code which escapes line brakes from https://github.com/mikel/mail/blob/02be6e37e850023421901042af3ff501e649d354/lib/mail/field.rb#L238


Davidslv:
@jeremy is there anything you need to make this PR merged and released on rubygems?

jeremy:
Discussion @ https://github.com/mikel/mail/issues/944#issuecomment-206119651. Backporting this commit is insufficient to fix the vuln and it breaks back compat to do it.

PR #951, opened 2016-01-08
Work with ruby 2.3's --enable-frozen-string-literal

jeremyevans:
These changes are the minimal ones necessary to allow Roda's specs
to pass.  There may well be other changes that are required.


jeremy:
:+1:

PR #952, opened 2016-01-11
Update mime types dependency

beanieboi:
hey,

this is an attempt to fix the missing pieces from #928 
i created my own PR because it was easier to work with.
what would be the best way to attribute @halostatue? i took most of the stuff from his PR #928 

let me know if there are any other issue to fix, i can contribute time to make this work.

thanks
ben

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,8 @@ gemfile:
 
 sudo: false
 
+bundler_args: --path vendor/bundle
+
 script: bundle exec rake
 
 env:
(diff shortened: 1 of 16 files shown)


jeremy:
Thanks @beanieboi. Rather than do a new commit with the same changes, start your work with the #928 branch. Add @halostatue's git repos as a remote, create a local branch from that, commit your new work, then submit a new PR. The `hub` utility provides some nice shortcuts for checking out/branching from a PR, too!

beanieboi:
ha! true! sorry, didn't think about this!

beanieboi:
new PR: https://github.com/mikel/mail/pull/953

Issue #955, opened 2016-01-14
Mail::Utilities.to_crlf Improvement


carsonreinke:
`Mail::Utilities.to_crlf` seems like it can be changed to not use a `Regexp` instead by using `String#encode`.

Running this test:
```ruby
require 'benchmark'

CRLF = "\r\n"
CRLF_REGEX = Regexp.new("(?<!\r)\n|\r(?!\n)")

string = "\n\r\n\n" * 10_000

Benchmark.bm do |x|
  x.report('to_crlf:') do
    10_000.times do
      string.gsub(CRLF_REGEX, CRLF)
    end
  end
  
  x.report('encode:') do
    10_000.times do
      string.encode(:universal_newline => true).encode!(:crlf_newline => true)
    end
  end
end
```
... I get these results on 2.3.0:
```
       user     system      total        real
to_crlf: 57.450000   0.120000  57.570000 ( 57.580348)
encode:  7.790000   0.090000   7.880000 (  7.889390)
```

Am I missing something here?  I will put in a PR if not.

kjg:
I'm not a maintainer or anything, but I say make a PR, include this perf info there. If all the tests still pass in the CI run for the PR, I can't see why they'd turn it down.

jeremy:
Please do pursue a PR! Note we support older rubies, so we'd need
version-specific shims.

jeremy:
:metal:

PR #956, opened 2016-01-14
Mail::Utilities.to_crlf/to_lf Improvement

carsonreinke:
See https://github.com/mikel/mail/issues/955.  Instead of `Regexp` use `String#encode`.

Running this test:
```ruby
require 'benchmark'

CRLF = "\r\n"
CRLF_REGEX = Regexp.new("(?<!\r)\n|\r(?!\n)")

string = "\n\r\n\n" * 10_000

Benchmark.bm do |x|
  x.report('regexp:') do
    10_000.times do
      string.gsub(CRLF_REGEX, CRLF)
    end
  end
  
  x.report('encode:') do
    10_000.times do
      string.encode(:universal_newline => true).encode!(:crlf_newline => true)
    end
  end
end
```
... I get these results:
```
       user     system      total        real
regexp: 57.450000   0.120000  57.570000 ( 57.580348)
encode:  7.790000   0.090000   7.880000 (  7.889390)
```

### DIFF
--- a/lib/mail/utilities.rb
+++ b/lib/mail/utilities.rb
@@ -223,7 +223,26 @@ module Mail
         enum.each_with_index.map(&block)
       end
 
-      def self.to_lf input
+    end
+    
+    # Test String#encode works correctly with line endings.
+    # Some versions of Ruby (e.g. MRI <1.9, JRuby, Rubinius) line ending
+    # normalization does not work correctly or did not have #encode.
+    if ("\r".encode(:universal_newline => true) rescue nil) == LF &&
+      (LF.encode(:crlf_newline => true) rescue nil) == CRLF
+      # Using String#encode is better performing than Regexp
+
+      def self.to_lf(input)
+        input.kind_of?(String) ? input.to_str.encode(:universal_newline => true) : ''
+      end
+
+      def self.to_crlf(input)
+        input.kind_of?(String) ? input.to_str.encode(:universal_newline => true).encode!(:crlf_newline => true) : ''
+      end
+
+    else
+
+      def self.to_lf(input)
         input.kind_of?(String) ? input.to_str.gsub(/\r\n|\r/, LF) : ''
       end
 
@@ -236,7 +255,7 @@ module Mail
         CRLF_REGEX = /\n|\r\n|\r/
       end
 
-      def self.to_crlf input
+      def self.to_crlf(input)
         input.kind_of?(String) ? input.to_str.gsub(CRLF_REGEX, CRLF) : ''
       end
 


jeremy:
JRuby and Rubinius aren't happy. Perhaps do feature detection (attempt an encode, rescue error) instead of a RUBY_VERSION check?

carsonreinke:
Looks like JRuby and Rubinius are not playing by the rules:

```
jruby-1.7.20 :001 > "\r".encode(:universal_newline => true)
 => "" 
```

```
rbx-2.5.2 :001 > "\n\n".encode(:crlf_newline => true)
 => "\n\n" 
```

The doc says ":universal_newline: Replaces CRLF (“rn”) and CR (“r”) with LF (“n”) if value is true" and ":crlf_newline: Replaces LF (“n”) with CRLF (“rn”) if value is true."

I could segment based on `RUBY_ENGINE`, but unsure if this is bad form.  Maybe just this: `if RUBY_VERSION >= '1.9' && RUBY_ENGINE == 'ruby'`.

jeremy:
How about doing a feature test: in the conditional, try the encode and
check that we get the expected result; if not, fall back to regexp.

carsonreinke:
Good idea.  CLI passed.

carsonreinke:
@jeremy Changed to use `to_str` because of https://github.com/mikel/mail/pull/281

lib/mail/utilities.rb
```diff
@@ -223,7 +223,26 @@ def map_with_index( enum, &block )
...
+    
+    # Test String#encode works correctly with line endings.
+    # Some versions of Ruby (e.g. MRI <1.9, JRuby, Rubinius) line ending
+    # normalization does not work correctly or did not have #encode.
+    if ("\r".encode(:universal_newline => true) rescue nil) == LF &&
+      (LF.encode(:crlf_newline => true) rescue nil) == CRLF
+      # Using String#encode is better performing than Regexp
+
+      def self.to_lf(input)
+        input.kind_of?(String) ? input.to_str.encode(:universal_newline => true) : ''
```
kjg: This form of String#encoding will encode the string to whatever Encoding.default_internal is set to. When running a rails app, Encoding.default_internal is set to UTF-8

This new behaviour of changing the encoding is breaking some specs in my rails app, it was expecting the string to still be encoding as ASCII-8bit at this point.

We can fix this by doing `input.to_str.encode(input.encoding, :universal_newline => true)`

I can make a PR.

carsonreinke: Opps, confirmed this behavior:

```
$ > Encoding.default_internal = Encoding::UTF_8
 => #<Encoding:UTF-8> 
$ > s = 'test'.encode(Encoding::ASCII_8BIT)
 => "test" 
$ > s.encoding
 => #<Encoding:ASCII-8BIT> 
$ > s.encode(:universal_newline => true).encoding
 => #<Encoding:UTF-8> 
```

PR #961, opened 2016-01-28
Fix encoding issue introduced by #956

kjg:
#956 changes the implementation of `Mail::Utilities#to_lf` and `#to_crlf` to use `String#encode` for better performance.

[The variant of String#encode used translates the string to `Encoding.default_internal`.
](http://ruby-doc.org/core-2.2.0/String.html#method-i-encode)
>The last form returns a copy of str transcoded to Encoding.default_internal.

In most rails apps this is set to "UTF-8". This can result in the encoding of the string being different than expected after being passed through these methods.

Instead, let's use the second form of `String#encode` in which we specify the encoding used is the same as the input string.

Benchmark shows no degradation in performance:
```ruby
CRLF = "\r\n"
CRLF_REGEX = Regexp.new("(?<!\r)\n|\r(?!\n)")

string = "\n\r\n\n" * 10_000

Benchmark.bm do |x|
  x.report('encode last form:') do
    10_000.times do
      string.encode(:universal_newline => true).encode!(:crlf_newline => true)
    end
  end

  x.report('encode second form:') do
    10_000.times do
      string.encode(string.encoding, :universal_newline => true).encode!(string.encoding, :crlf_newline => true)
    end
  end
end
```

results:
```
       user     system      total        real
encode last form:  9.240000   0.410000   9.650000 (  9.706578)
encode second form:  8.950000   0.230000   9.180000 (  9.218603)
```

### DIFF
--- a/lib/mail/utilities.rb
+++ b/lib/mail/utilities.rb
@@ -233,11 +233,11 @@ module Mail
       # Using String#encode is better performing than Regexp
 
       def self.to_lf(input)
-        input.kind_of?(String) ? input.to_str.encode(:universal_newline => true) : ''
+        input.kind_of?(String) ? input.to_str.encode(input.encoding, :universal_newline => true) : ''
       end
 
       def self.to_crlf(input)
-        input.kind_of?(String) ? input.to_str.encode(:universal_newline => true).encode!(:crlf_newline => true) : ''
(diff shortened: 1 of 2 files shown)


jeremy:
Nice catch! Thanks @kjg

Issue #965, opened 2016-02-08
Problem with frozen strings


aried3r:
Even after #951 I have problems using this gem because of [this method](https://github.com/mikel/mail/blob/728942f2bc3a24fc84d6653e80d842b08a8858a0/lib/mail/utilities.rb#L23-L37) at least. The problem is `.force_encoding` which tries to modify `self`. So I guess we should `.dup` the string before modifying?

Stack trace calling `.format` on `Mail::Address`
```
/usr/local/opt/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/mail-2.6.3/lib/mail/utilities.rb:22:in `force_encoding': can't modify frozen String (RuntimeError)
	from /usr/local/opt/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/mail-2.6.3/lib/mail/utilities.rb:22:in `quote_phrase'
	from /usr/local/opt/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/mail-2.6.3/lib/mail/elements/address.rb:51:in `format'
```

jeremy:
+1 to not mutating its argument. Could you write up the change + a test?

aried3r:
Fixed in 2.6.4. Thanks!

MaxPleaner:
I'm still getting `*** RuntimeError Exception: can't modify frozen String` for any of the commands to fetch email from a Gmail account. e.g. `Mail.first` or `Mail.all`.

I have in my Gemfile.lock mail (2.6.6) and Ruby 2.5.0-dev

I never added the frozen string literal: true comment either, by the way.

jeremy:
@MaxPleaner Do you have a full backtrace for the exception? Thanks!

sungsong88:
@jeremy I have the same issue... How can I grab a backtrace please? I'll post it here.

PR #970, opened 2016-03-03
use frozen string literal for ruby 2.3.0+

twalpole:


lib/mail/elements/content_location_element.rb
```diff
@@ -1,21 +1,22 @@
...
     def location
       @location
     end
-    
+
     def to_s(*args)
       location.to_s
     end
-    
+
```
jeremy: Superfluous whitespace?

lib/mail/fields/content_disposition_field.rb
```diff
@@ -9,6 +10,7 @@ class ContentDispositionField < StructuredField
 
     def initialize(value = nil, charset = 'utf-8')
       self.charset = charset
+      value = value.dup if value.is_a?(String) && value.frozen?
       ensure_filename_quoted(value)
       super(CAPITALIZED_FIELD, strip_field(FIELD_NAME, value), charset)
       self.parse
```
jeremy: Perhaps `#ensure_filename_quoted` shouldn't be modifying `value` in-place, instead?

lib/mail/fields/unstructured_field.rb
```diff
@@ -127,7 +128,7 @@ def fold(prepend = 0) # :nodoc:
           if first
             first = !first
           else
-            word = " " << word
+            word = " #{word}"
```
jeremy: 👍

lib/mail/utilities.rb
```diff
@@ -224,7 +225,7 @@ def map_with_index( enum, &block )
       end
 
     end
-    
+
```
jeremy: Whitespace

spec/mail/elements/date_time_element_spec.rb
```diff
@@ -1,3 +1,5 @@
+# encoding: utf-8
```
jeremy: Is the explicit encoding declaration necessary?

spec/mail/encodings_spec.rb
```diff
@@ -837,7 +838,7 @@
   describe ".charset_encoder" do
     class CustomEncoder
       def encode(str, charset)
-        "#{str}-#{charset}"
+        String.new("#{str}-#{charset}")
```
jeremy: Does this need to be a non-frozen string?

spec/mail/fields/common/common_field_spec.rb
```diff
@@ -16,15 +17,15 @@
...
     it "should leave ascii alone" do
       field = Mail::SubjectField.new("This is a test")
       expect(field.encoded).to eq "Subject: This is a test\r\n"
       expect(field.decoded).to eq "This is a test"
     end
-    
+
     it "should encode a utf-8 string as utf-8 quoted printable" do
-      value = "かきくけこ"
+      value = String.new("かきくけこ")
```
jeremy: Why does this value need to be mutable?

spec/mail/fields/unstructured_field_spec.rb
```diff
@@ -150,7 +151,7 @@
 
     it "should fold itself if it is non us-ascii" do
       @original = $KCODE if RUBY_VERSION < '1.9'
-      string = "This is あ really long string This is あ really long string This is あ really long string This is あ really long string This is あ really long string"
+      string = String.new("This is あ really long string This is あ really long string This is あ really long string This is あ really long string This is あ really long string")
       @field = Mail::UnstructuredField.new("Subject", string)
       if string.respond_to?(:force_encoding)
         string = string.force_encoding('UTF-8')
```
jeremy: Would `#dup` at this point rather than doing `String.new`

spec/mail/fields/unstructured_field_spec.rb
```diff
@@ -165,7 +166,7 @@
...
+      string = String.new(%|{"unique_args": {"mailing_id":147,"account_id":2}, "to": ["larspind@gmail.com"], "category": "mailing", "filters": {"domainkeys": {"settings": {"domain":1,"enable":1}}}, "sub": {"{{open_image_url}}": ["http://betaling.larspind.local/O/token/147/Mailing::FakeRecipient"], "{{name}}": ["[FIRST NAME]"], "{{signup_reminder}}": ["(her kommer til at stå hvornår folk har skrevet sig op ...)"], "{{unsubscribe_url}}": ["http://betaling.larspind.local/U/token/147/Mailing::FakeRecipient"], "{{email}}": ["larspind@gmail.com"], "{{link:308}}": ["http://betaling.larspind.local/L/308/0/Mailing::FakeRecipient"], "{{confirm_url}}": [""], "{{ref}}": ["[REF]"]}}|)
       @field = Mail::UnstructuredField.new("X-SMTPAPI", string)
       if string.respond_to?(:force_encoding)
         string = string.force_encoding('UTF-8')
```
jeremy: Ditto

spec/mail/header_spec.rb
```diff
@@ -557,13 +558,13 @@
       expect(@traced_header.fields[5].info).to eq "from agw2 by xxx.xxxx.xxx"
       expect(@traced_header.fields[6].field.class).to eq Mail::ToField
     end
-    
+
```
jeremy: Whitespace

spec/mail/utilities_spec.rb
```diff
@@ -403,7 +404,7 @@
     it "should say an empty string is blank" do
       expect(Mail::Utilities.blank?('')).to be_truthy
       expect(Mail::Utilities.blank?(" ")).to be_truthy
-      a = ''; 1000.times { a << ' ' }
+      a = String.new(''); 1000.times { a << ' ' }
```
jeremy:
```ruby
a = ' ' * 1000
```

spec/spec_helper.rb
```diff
@@ -47,7 +48,7 @@ def fixture(*name)
 # of the letters and numbers work.
 def ascii(from = 33, to = 126)
   chars = []
-  from.upto(to) { |c| chars << ('' << c) }
+  from.upto(to) { |c| chars << (String.new('') << c) }
```
jeremy:
```ruby
chars = (from..to).map(&:chr)
```

jeremy:
Nice work, @twalpole! Good call including the pragma everywhere.

Some of these cases suggest that methods are mutating strings that they probably shouldn't.

`String.new(…)` makes a copy of the argument, so selective use of `#dup` may end up feeling cleaner. Having to shed our literal strings in so many places frequently feels like prematurely avoiding frozen strings.

lib/mail/elements/content_location_element.rb
twalpole: The change removes the superfluous whitespace that was in the file (my editor did that I guess) -- if you're saying the line is unneeded, it was there before but with some spaces at the beginning

lib/mail/fields/content_disposition_field.rb
twalpole: I'm fine with changing that if you're ok with me changing that behavior ?- it is a private method so it's not public API

lib/mail/utilities.rb
twalpole: same as the first -- the change actually removed whitespace

spec/mail/elements/date_time_element_spec.rb
twalpole: no - not sure why that ended up in there -- leftover from debugging something I guess -- will remove

spec/mail/encodings_spec.rb
twalpole: yes - unless Ruby19.q_value_decode is changed to not call force_encoding on the result of it

spec/mail/fields/common/common_field_spec.rb
twalpole: because force_encoding is called on it at line 33

spec/mail/fields/unstructured_field_spec.rb
twalpole: Ditto

spec/mail/header_spec.rb
twalpole: same as above

spec/mail/utilities_spec.rb
twalpole: ok -- but not sure this change really has anything to do with a PR thats about frozen string literals

spec/spec_helper.rb
twalpole: ok -- but not sure this change really has anything to do with a PR thats about frozen string literals

spec/mail/fields/unstructured_field_spec.rb
twalpole: calling #dup moves it off to the right end of a really long string and immediately clear that it's not a frozen literal - can change if really wanted

lib/mail/fields/common/common_field.rb
```diff
@@ -51,7 +52,9 @@ def strip_field(field_name, value)
     FILENAME_RE = /\b(filename|name)=([^;"\r\n]+\s[^;"\r\n]+)/
     def ensure_filename_quoted(value)
       if value.is_a?(String)
-        value.sub! FILENAME_RE, '\1="\2"'
+        value.sub FILENAME_RE, '\1="\2"'
+      else
+        value
```
jeremy: 👍

lib/mail/utilities.rb
jeremy: Yep yep. These are OK, but they're unrelated to the PR so they add a bit of noise for review and to the git history, where a `git blame` will show this commit that has nothing to do with the subject code.

spec/mail/encoding_spec.rb
```diff
@@ -115,13 +116,14 @@
...
+      subject = String.new("This is あ string")
       subject.force_encoding('ISO8859-1') if RUBY_VERSION > '1.9'
       mail.subject = subject
       result = mail[:subject].encoded
-      string = "Subject: =?ISO-8859-1?Q?This_is_=E3=81=82_string?=\r\n"
+      string = String.new("Subject: =?ISO-8859-1?Q?This_is_=E3=81=82_string?=\r\n")
       if RUBY_VERSION > '1.9'
         string.force_encoding('ISO8859-1')
+        result = result.dup if result.frozen?
         result.force_encoding('ISO8859-1')
```
jeremy: Having to use `String.new` everywhere is feeling off.

How about using literal string syntax, instead, and duping as needed:
```ruby
  string = string.dup.force_encoding('ISO8859-1')
  result = result.dup.force_encoding('ISO8859-1')
```

Then we can use standard syntax up-front and skip the `#frozen?` checks.

spec/mail/encodings_spec.rb
jeremy: 👍

spec/mail/fields/common/common_field_spec.rb
jeremy: Hmm… we shouldn't need to force-encode to utf-8 since the source encoding is already utf-8, huh?

spec/mail/fields/unstructured_field_spec.rb
jeremy: I think we need to work to keep the literal string syntax where we can. We can do `string = string.dup.force_encoding(…)` in many of these cases.

However, this is another case where the `#force_encoding` look unneeded since the source encoding is already utf-8.

spec/mail/fields/unstructured_field_spec.rb
jeremy: 😊

spec/mail/fields/common/common_field_spec.rb
twalpole: I assumed maybe 1.9 didn't obey the encoding or something -- can't remember off the top of my head what versions supported what.  Since it was in the test already I wasn't interested in removing it and if it can be removed it should probably be in another PR

twalpole:
@jeremy Ok -I've swapped to dup in most places and also cleaned up the raise_error warnings to make sure they weren't hiding frozen string errors.

spec/mail/fields/common/common_field_spec.rb
jeremy: Changing in another PR works too.

Magic comments were added in 1.9.0: https://github.com/ruby/ruby/commit/b1e97c9070e61ce4e7f1be0a02cef9ef302bba8f

lib/mail/attachments_list.rb
```diff
@@ -71,6 +72,7 @@ def []=(name, value)
 
       if hash[:body].respond_to? :force_encoding and hash[:body].respond_to? :valid_encoding?
         if not hash[:body].valid_encoding? and default_values[:content_transfer_encoding].downcase == "binary"
+          hash[:body] = hash[:body].dup if hash[:body].frozen?
           hash[:body].force_encoding("BINARY")
```
jeremy: Another spot to sidestep the frozen check:
```ruby
hash[:body] = hash[:body].dup.force_encoding("BINARY")
```

jeremy:
Looking good @twalpole!

Would you squash down to a single commit?

lib/mail/attachments_list.rb
twalpole: This is in the gems code as opposed to test, and I figured the body could be large -- for performance and memory reasons I didn't want to dup it unless necessary

twalpole:
@jeremy done

aried3r:
Oh, that's great!

I guess that fixes #965 except the method I mentioned there also uses `.force_encoding` but was not modified in this PR. Is the input of the method now already "sanitized"?

twalpole:
@aried3r  The one string `quote_phrase` is called with in the code (display_name) is now mutable so from that perspective it's fixed - If using the method from outside it still forces the encoding on the string and then forces it back to the original encoding - I'll do a quick PR for that

PR #978, opened 2016-03-31
Handle unknown b-value encoding

kjg:
by treating it as ascii as well as ignoring any invalid chars

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -1,5 +1,8 @@
 == HEAD
 
+Bugs:
+* #978 - Fix for invalid chars being left in a string for invalid b_value from encoding (kjg)
+
 == Version 2.6.4 - Wed Mar 23 08:16 -0700 2016 Jeremy Daer <jeremydaer@gmail.com>
 
 Features:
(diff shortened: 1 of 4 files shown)


kjg:
Quick question: How would you feel about replacing invalid chars with `"�"` instead of `""`. That way it is clear that something went "wrong" during translation where we otherwise just pretend everything was okay with no notice.

jeremy:
Good call on � 👍

kjg:
Should I change that in this PR or a new one?

Any idea offhand how to get Iconv to do � in 1.8.7?

If it isn’t easy / built in to Iconv, is it worth it to add for 1.8.7, or should I do it just for the String#encode versions?


On March 31, 2016 at 11:15:25 AM, Jeremy Daer (notifications@github.com) wrote:

Good call on � 👍

kjg:
Yeah, as far as I can tell, There is no good way to specify a replacement character in 1.8.7. Any 1.8.7 solution that I can think of to replace invalid characters would involve iterating over bytes and be really messy.

So the question becomes do we scrap � and leave the replacement as "" across the board, or do we allow the 1.8.7 output to be different.

I vote for allowing the 1.8.7 output to be slightly different.

bf4:
@kjg Maybe https://github.com/bf4/encoded_string/blob/master/lib/encoded_string.rb#L14-L19 helps? I don't have 1.8.7 to test

jeremy:
Separate PR makes sense.

kjg:
This is passing now after a rebase. I'll make the replacement character change in a separate PR

Issue #980, opened 2016-04-05
Unfolding of subject header removes consecutive spaces


aselder:
``` ruby
s = "Your Order  with ABC"
Mail::Field.new('subject', s, 'UTF-8')
f.to_s == s # => false
f.to_s # => "Your Order with ABC"
```

I've narrowed it down to line 236 in field.rb:

``` ruby
s.gsub(/[\r\n \t]+/m, ' ') # => "Your Order with ABC"
```

Is there any reason to strip turn multiple spaces in a subject field. If nothing else, this violates the principle of least surprise.

I was writing a spec to test a mailer and the subject I sent in wasn't what I got back out.

jeremy:
Agree it's odd. What does git history indicate, @aselder?

aselder:
@jeremy This was last changed 2/1/13 in 72befdc4dab3e6e288ce226a7da2aa474cf5be83, with the comment "This has little effect on the specs, but on my header reading example it
makes about a 10x performance difference"

So it looks like this was a performance enhancement that led to a regression. The only reason I'm discovering it is we're upgrade from Rails 3 (mail 2.5.x) to Rails 4 (mail 2.6.x). All of a sudden a bunch of specs started failing.

aselder:
The comment says "Unfolding is accomplished by simply removing any CRLF that is immediately followed by WSP."

How ever this is not what the regex does. Based on the comment, I can probably correct the regex.

vihai:
This issue has major consequences when it comes to S/MIME signing the messages as the headers contents are not preserved and the signature is made invalid.

jeremy:
Fixed by #982

PR #981, opened 2016-04-05
Handle whitespace better when unfolding subject.

aselder:
When unfolding a subject only remove whitespace on either side of a CRLF.

Multiple consecutive spaces in a subject could have meaning to user of the library.

This also fixes a change in behavior from Mail 2.5.x to Mail 2.6.x. In Mail 2.5.x, consecutive simple spaces were not touched.

This is a fix for issue #980

### DIFF
--- a/lib/mail/field.rb
+++ b/lib/mail/field.rb
@@ -232,12 +232,13 @@ module Mail
     #  that is immediately followed by WSP.  Each header field should be
     #  treated in its unfolded form for further syntactic and semantic
     #  evaluation.
-    def unfold(string)
-      string.gsub(/[\r\n \t]+/m, ' ')
+    def unfold(name, value)
+      regex = name == 'Subject' ? /\s*\r\n\s+/m : /[\r\n \t]+/m
+      value.gsub(regex, ' ')
     end
 
     def create_field(name, value, charset)
-      value = unfold(value) if value.is_a?(String)
+      value = unfold(name, value) if value.is_a?(String)
 
       begin
         new_field(name, value, charset)
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -2,6 +2,7 @@
 
 Bugs:
 * #978 - Fix for invalid chars being left in a string for invalid b_value from encoding (kjg)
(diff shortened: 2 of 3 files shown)


lib/mail/field.rb
```diff
@@ -232,12 +232,13 @@ def split(raw_field)
     #  that is immediately followed by WSP.  Each header field should be
     #  treated in its unfolded form for further syntactic and semantic
     #  evaluation.
-    def unfold(string)
-      string.gsub(/[\r\n \t]+/m, ' ')
+    def unfold(name, value)
+      regex = name == 'Subject' ? /\s*\r\n\s+/m : /[\r\n \t]+/m
```
kjg: This feels a little inelegant to me. Is there anyway to just modify the original regex?

aselder: @kjg That was my original approach, but based on existing specs, there are some header fields where the whitespace is supposed to be stripped out. I wasn't able to come up with a single regex that passed both my new specs and all existing specs.

kjg: Which fields? Is there a chance the specs are wrong too and written just to
match the existing behavior?

aselder: @kjg. That's quite possible, that the specs are wrong. But I wanted to cause as little disruption as possible.

An example it the 'To' header, in the spec right above the 2 I added:

```ruby
    it "should remove multiple spaces during unfolding a header" do
      header = Mail::Header.new("To: Mikel,\r\n   Lindsaar,     Bob")
      expect(header['To'].value).to eq 'Mikel, Lindsaar, Bob'
    end
```

aselder:
@kjg I'm happy to consolidate to a single regex and updating the specs, it that's what you guys would prefer.

lib/mail/field.rb
kjg: I'd expect the "unfolding" there to only remove the spaces after the \r\n
and before Lindsaar. Is there an rfc that specifies this case?

aselder: @kjg let me look and i'll get back to you

jeremy:
Nice work! I took a look and corrected the unfolding scheme to preserve whitespace, including correcting the specs: #982

PR #982, opened 2016-04-06
Preserve unfolded whitespace

### LINKED ISSUE #980: Unfolding of subject header removes consecutive spaces

jeremy:
We've long squashed repeated whitespace (space and horizontal tab) to a single space when we unfold headers. We're supposed to just remove the CRLF and leave the whitespace as-is.

This is a breaking parsing change that will leave some users with extra whitespace that was silently squished before.

Think we have some similar issues with our header folding.

Fixes #980.

### DIFF
--- a/spec/mail/header_spec.rb
+++ b/spec/mail/header_spec.rb
@@ -431,9 +431,9 @@ describe Mail::Header do
       expect(header['To'].value).to eq 'Mikel, Lindsaar, Bob'
     end
 
-    it "should remove multiple spaces during unfolding a header" do
-      header = Mail::Header.new("To: Mikel,\r\n   Lindsaar,     Bob")
-      expect(header['To'].value).to eq 'Mikel, Lindsaar, Bob'
+    it "should preserve whitespace when unfolding a header" do
+      header = Mail::Header.new("To: Mikel,\r\n\t   Lindsaar,     Bob")
+      expect(header['To'].value).to eq "Mikel,\t   Lindsaar,     Bob"
     end
 
     it "should handle a crazy long folded header" do
@@ -445,7 +445,7 @@ Received: from [127.0.220.158] (helo=fg-out-1718.google.com)
 	for support@aaa.somewhere.com; Thu, 05 Jun 2008 10:53:29 -0700
 HERE
       header = Mail::Header.new(header_text.gsub(/\n/, "\r\n"))
-      expect(header['Received'].value).to eq 'from [127.0.220.158] (helo=fg-out-1718.google.com) by smtp.totallyrandom.com with esmtp (Exim 4.68) (envelope-from <stuff+caf_=support=aaa.somewhere.com@gmail.com>) id 1K4JeQ-0005Nd-Ij for support@aaa.somewhere.com; Thu, 05 Jun 2008 10:53:29 -0700'
+      expect(header['Received'].value).to eq "from [127.0.220.158] (helo=fg-out-1718.google.com)\tby smtp.totallyrandom.com with esmtp (Exim 4.68)\t(envelope-from <stuff+caf_=support=aaa.somewhere.com@gmail.com>)\tid 1K4JeQ-0005Nd-Ij\tfor support@aaa.somewhere.com; Thu, 05 Jun 2008 10:53:29 -0700"
     end
 
     it "should convert all lonesome LFs to CRLF" do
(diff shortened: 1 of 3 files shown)


jeremy:
Note we originally did this unfolding: https://github.com/mikel/mail/commit/773143be5f4b5f20c34196b414b38c4571086d28#diff-66f4c85e7675fea548f6ae0c81b9de04R76
```ruby
def unfold(string)
  string.gsub(/\n([\t\s]+)/, '\1')
end
```

The behavior change is https://github.com/mikel/mail/commit/fdfa9a1d85a447b94c84828f8dbb5e35fdc3cfa6#diff-66f4c85e7675fea548f6ae0c81b9de04R189
which includes lots of other work; unclear why.

kjg:
Does / will something else strip the tabs from the beginning of subsequent lines that are due to being folded?

jeremy:
@kjg Depends! The subject field is an unstructured field where all non-CRLF whitespace is treated as literal. So these tabs are correct.

Structured fields have [folding white space](https://tools.ietf.org/html/rfc2822#section-3.2.3) (distinct from [header folding](https://tools.ietf.org/html/rfc2822#section-2.2.3)) which treats runs of whitespace/CRLF/whitespace as semantically equivalent to a single space.

spec/mail/header_spec.rb
```diff
@@ -564,7 +564,7 @@
   describe "encoding" do
     it "should output a parsed version of itself to US-ASCII on encoded and tidy up and sort correctly" do
       encoded = Mail::Header.new("To: Mikel\r\n\sLindsaar <mikel@test.lindsaar.net>\r\nFrom: bob\r\n\s<bob@test.lindsaar.net>\r\nSubject: This is\r\n a long\r\n\s \t \t \t    badly formatted             \r\n       \t\t  \t       field").encoded
-      result = "From: bob <bob@test.lindsaar.net>\r\nTo: Mikel Lindsaar <mikel@test.lindsaar.net>\r\nSubject: This is a long badly formatted field\r\n"
+      result = "From: bob <bob@test.lindsaar.net>\r\nTo: Mikel Lindsaar <mikel@test.lindsaar.net>\r\nSubject: This is a long           badly formatted                             \r\n   field\r\n"
```
jeremy: Note this is apparently incorrect, too. The subject should have preserved its whitespace when it's folded.

kjg:
@jeremy Thanks for explaining that! I didn't realize there was a different in folding between Structured and Unstructured fields. I notice that the [ReceivedField](https://github.com/mikel/mail/blob/df48a05a7fb5a4271e6df12da7afb26a53494a18/lib/mail/fields/received_field.rb#L24) is a StructuredField. Does that mean `expect(mail.received[0].info).to eq 'from x.y.test by e` etc. should stay with the extra white space removed?

kjg:
If I'm reading
```
name-val-list   =       [CFWS] [name-val-pair *(CFWS name-val-pair)]

name-val-pair   =       item-name CFWS item-value
```
right from https://tools.ietf.org/html/rfc2822#section-3.6.7

It looks like the original spec https://github.com/mikel/mail/pull/982/files#diff-23e3ce655ed29609c6d8b144874773d7L158 was valid here where it removed the white space.

Are you getting the same reading of the rfc?

spec/mail/header_spec.rb
aselder: I think the CRLF needs to removed from the subject field.Otherwise I think the SMTP server will be expecting a new field to start on the next line.

jeremy:
@kjg Our parser doesn't handle Received correctly. It parses the whole value into "info" as if it's an unstructured field. It'd need to parse the name-val-list instead.

spec/mail/header_spec.rb
jeremy: Yep. It's definitely wrong.

Issue #963, opened 2016-02-01
NoMethodError in get_display_name


amitsaharana:
If Mail::Address is initialized with empty/nil email then accessing .display_name gives:
NoMethodError: undefined method `display_name' for nil:NilClass in 
lib/mail/elements/address.rb:207:in `get_display_name'
due to @data being empty/nil. This is not consistent with address and other accessors, which return nil.

PR #983, opened 2016-04-07
Allow display name to be nil by default or to be reset

carsonreinke:
Fixes https://github.com/mikel/mail/issues/963


jeremy:
Thanks @carsonreinke! [Backported to 2-6-stable](0cd5757217e668f614e7536062c0e54ad5c9c847) also.

marcandre:
Got bitten by that too. Any way to get an official release?

PR #986, opened 2016-04-15
fix issue calling .to_s multiple times corrupts message body

robguthrie:
I have discovered a bug with calling message.to_s when the body of the message contains non-latin characters.

I first reported the error in [ankane/ahoy_email's issues](https://github.com/ankane/ahoy_email/issues/50) where I detailed the buggy behaviour.

I've created a work around but I'm not sure if it's the best approach. Anyway it works for me. Thanks for your time.

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -1815,11 +1815,15 @@ module Mail
     # all headers, attachments, etc.  This is an encoded email in US-ASCII,
     # so it is able to be directly sent to an email server.
     def encoded
-      ready_to_send!
-      buffer = header.encoded
-      buffer << "\r\n"
-      buffer << body.encoded(content_transfer_encoding)
-      buffer
+      if @buffer
+        @buffer
+      else
+        ready_to_send!
+        @buffer = header.encoded
+        @buffer << "\r\n"
+        @buffer << body.encoded(content_transfer_encoding)
+        @buffer
+      end
     end
 
     def without_attachments!


kjg:
Can you please also include an illustrative test case that fails without the change and passes with the change?

lib/mail/message.rb
```diff
@@ -1815,11 +1815,15 @@ def encode!
...
-      buffer
+      if @buffer
+        @buffer
+      else
+        ready_to_send!
+        @buffer = header.encoded
+        @buffer << "\r\n"
+        @buffer << body.encoded(content_transfer_encoding)
+        @buffer
+      end
```
jeremy: Rather than memoize the encoded message (which means any change to the message won't be reflected in subsequence `#encoded` calls), let's track down what's mutating your message body.

Could you debug further to identify? Perhaps you're getting double transfer decoding?

robguthrie:
Thanks for the prompt.

I'm not quite sure how to best go about creating a test.
I'm using ActionMailer to generate an email, and if message.to_s is called twice this behaviuor occurs.

How should I go about setting up a test case within mail?

Thanks for your help!

jeremy:
@robguthrie Think of how you'd create a test case within your app. Then start trimming it down to a case that uses the Mail lib only (create a Mail.new, set the body, etc) until you have a minimal case that reproduces the issue you're seeing. Then find a spot in the Mail specs that seems related to mail body encoding and translate your test over to a mail spec 😁

robguthrie:
I'm having a really tough time reproducing the problem. 

I can still reproduce it within my app, but removing all the rails supplied stuff, and trying to understand how I should recreate the message within the test environment is taking me too much time.

I wonder if there is a test case where someone composes an email just as rails would?

jeremy:
> but removing all the rails supplied stuff, and trying to understand how I should recreate the message

Check out the Action Mailer source code and test cases to see this. It uses the Mail lib internally to construct the message 👍

thatguysimon:
Hey @robguthrie, just encountered this same issue myself, any follow up? Are you just using your workaround?

robguthrie:
@thatguysimon nope, since then another developer dug a bit deeper and patched the problem in premailer, so we use a fork of that.

https://github.com/gdpelican/premailer-rails

thatguysimon:
Premailer? I'm confused.. the workaround you suggested in this pull request did stop premailer from working when I tried it, but I didn't think the problem stems from premailer.
Either way, this seems like an important issue, why not merge the patch into the original premailer?

robguthrie:
The bug is confusing. We've not fixed it properly, just for our needs.

gdpelican:
The bug existing in premailer-rails and was patched in premailer-rails 1.9.4

Check out [here](https://github.com/fphilipe/premailer-rails/pull/166) for more info.

thatguysimon:
Awesome, updating premailer to 1.9.4 did the trick. Thanks guys!
I'd say this pull request can be closed.

PR #987, opened 2016-04-20
Sort attachments to end rebased

### LINKED ISSUE #655: Text attachments appear as message body in some mail clients

ahorek:
based on https://github.com/mikel/mail/pull/675 Fixes #655


jeremy:
Will keep #675 and resolve when merged. Thank you @ahorek 😊

npickens:
Merge conflict has been resolved for #675.

PR #996, opened 2016-05-04
Don't return the delivery_status_part on non report emails

kjg:
Currently `Mail::Message#bounced?` will return `true` when someone happens to forward a bounce email resulting in the forwarded email properly being a multipart/mixed, but with a `message/delivery-status` part

Mail#bounced? should only be true if the top level message is a multipart/report. The same should hold true for all the multipart/report convenience methods.


kjg:
@jeremy Can you please take a look at this one too?

jeremy:
Squashed 8130b9c216bfb3149e10aef294696400da0b0954

PR #998, opened 2016-05-10
Allow blank charset or lang param hash

kjg:
Allow for blank charset or lang in param hash

as specified in [RFC 2231](https://tools.ietf.org/html/rfc2231#section-4)

"Note that it is perfectly permissible to leave either the character
   set or language field blank.  Note also that the single quote
   delimiters MUST be present even when one of the field values is
   omitted.  This is done when either character set, language, or both
   are not relevant to the parameter value at hand."

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -2,6 +2,7 @@
 
 Bugs:
 * #978 - Fix for invalid chars being left in a string for invalid b_value from encoding (kjg)
+* #998 - Fix for parsing of encoded header parameters (such as attachment names) containing a blank locale (kjg)
 
 == Version 2.6.4 - Wed Mar 23 08:16 -0700 2016 Jeremy Daer <jeremydaer@gmail.com>
 
--- a/spec/mail/fields/common/parameter_hash_spec.rb
+++ b/spec/mail/fields/common/parameter_hash_spec.rb
@@ -51,6 +51,24 @@ describe Mail::ParameterHash do
     expect(hash['value']).to eq "This is even more ***fun*** isn't it"
   end
 
+  it "allows for blank language" do
+    hash = Mail::ParameterHash.new
+    hash.merge!({'value*' => "us-ascii''Hello%20there"})
+    expect(hash['value']).to eq "Hello there"
+  end
+
+  it "allows for blank charset" do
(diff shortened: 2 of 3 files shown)


kjg:
@jeremy I'd love it if you have time to take a look at this one real quick. This one is harder for me to code around in my own app.

Thanks!

jeremy:
Looks like a lossy-transcode commit snuck in here too—a topic for a separate PR.

CHANGELOG.rdoc
```diff
@@ -2,6 +2,7 @@
 
 Bugs:
 * #978 - Fix for invalid chars being left in a string for invalid b_value from encoding (kjg)
+* #998 - ParameterHash fixes to allow for blanks and to properly transcode the values (kjg)
```
jeremy: Users don't know what `ParameterHash` is or how this is relevant to their apps. What messages can now be parsed that couldn't before?

kjg:
That commit is intentionally in this PR as I believe it is a bug as well,
but I can make it separate if you'd like.

kjg:
I removed the transcode now. Thanks

kjg:
I move the transcode to https://github.com/mikel/mail/pull/1000. Thanks!

CHANGELOG.rdoc
kjg: I've now improved the CHANGELOG wording

spec/mail/fields/common/parameter_hash_spec.rb
```diff
@@ -51,6 +51,18 @@
...
+    hash.merge!({'value*' => "us-ascii''Hello%20there"})
+    expect(hash['value']).to eq "Hello there"
+  end
+
+  it "should allow for blank char set" do
+    hash = Mail::ParameterHash.new
+    hash.merge!({'value*' => "'en'Hello%20there"})
+    expect(hash['value']).to eq "Hello there"
+  end
+
```
jeremy: Think we should test for both blank, too?

kjg: I added a test for both blank

jeremy:
Squashed 6fa8179c3806d52ff9ef8b308fe9c4dbbb4ff3da

PR #1000, opened 2016-05-10
Update param_decode to actually transcode to utf-8

kjg:


lib/mail/version_specific/ruby_1_9.rb
```diff
@@ -125,7 +125,11 @@ def Ruby19.q_value_decode(str)
     def Ruby19.param_decode(str, encoding)
       str = uri_parser.unescape(str)
       str = charset_encoder.encode(str, encoding) if encoding
-      str
+      decoded = str.encode(Encoding::UTF_8, :invalid => :replace, :replace => "")
```
jeremy: Let's use the [Unicode replacement char](https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character) rather than drop the character entirely.

lib/mail/version_specific/ruby_1_9.rb
```diff
@@ -125,7 +125,11 @@ def Ruby19.q_value_decode(str)
...
+      decoded.valid_encoding? ? decoded : decoded.encode(Encoding::UTF_16LE, :invalid => :replace, :replace => "").encode(Encoding::UTF_8)
```
jeremy: When will `decoded.valid_encoding?` be false here if we just did a lossy transcode to utf-8?

lib/mail/version_specific/ruby_1_9.rb
```diff
@@ -125,7 +125,11 @@ def Ruby19.q_value_decode(str)
...
+    rescue Encoding::UndefinedConversionError, ArgumentError, Encoding::ConverterNotFoundError
```
jeremy: Do we have test coverage for these cases?

Should we swallow these errors and return a utf-8 string with invalid encoding?

lib/mail/version_specific/ruby_1_9.rb
```diff
@@ -125,7 +125,11 @@ def Ruby19.q_value_decode(str)
...
+      warn "Encoding conversion failed #{$!}"
```
jeremy: This will show up as noise in stderr and logs—no way to trace back to its cause.

jeremy:
This changes the contract of `param_decode`. It attempts to respect the charset that's specified and map it to the corresponding Ruby string encoding. Now it transcodes from the specified charset to utf-8. Is this desirable for non-utf-8 values? Can we round-trip multilingual values?

lib/mail/version_specific/ruby_1_9.rb
kjg: I copied the strategy used in the two other _decode methods with another PR coming to refactor this into a shared method that all three will use.

Should I make that other PR first?

lib/mail/version_specific/ruby_1_9.rb
kjg: Yes, yes it will. And it annoys me every time it shows up in my tests. Should this instead add to Mail.errors or something?

Do you want me to do this before or after refactoring the conversion to UTF-8 the other _decode methods use

kjg:
Yes, it does effectively change the contract of `param_decode`, however I believe the current behaviour is unexpected. The method performs a very similar function and is used in the same context as q and b value decode. Both of those methods return the value transcoded from the parsed encoding specified into UTF-8.

`param_decode` is the method that ultimately produces the value returned by `part.filename` for example. All other similar parameter lookups like `message.subject` for example always return the value in UTF-8 and I think it is unexpected for `part.filename` to behave differently.

lib/mail/version_specific/ruby_1_9.rb
kjg: I agree, I remember that we talked about this in https://github.com/mikel/mail/pull/978

My end goal is that `b_value_decode` `q_value_decode` and `param_decode` all use the same method to encode to UTF-8 and that they all use the unicode replacement car. We can get there using whatever path you'd like, but my idea was to:

1. add this to param_decode first
2. refactor the three implementations down to 1 method
3. change that one method to use the unicode replacement char

How do you feel about that approach? Do you have something different in mind?

lib/mail/version_specific/ruby_1_9.rb
kjg: In ruby 1.9.3 "utf-8" to "utf-8" encoding is a no-op:

```ruby
[ ruby-1.9.3-p547 ][ mail:007:0 ] >> invalid_char = "\x8A"
=> "\x8A"
[ ruby-1.9.3-p547 ][ mail:008:0 ] >> invalid_char.encoding
=> #<Encoding:UTF-8>
[ ruby-1.9.3-p547 ][ mail:009:0 ] >> invalid_char.valid_encoding?
=> false
[ ruby-1.9.3-p547 ][ mail:012:0 ] >> converted = invalid_char.encode(Encoding::UTF_8, :undef => :replace, :invalid => :replace, :replace => "")
=> "\x8A"
[ ruby-1.9.3-p547 ][ mail:013:0 ] >> converted.valid_encoding?
=> false
```

The conversion to utf_16 and back is a workaround for that no-op

kjg:
@jeremy Any guidance on how you'd like me to proceed here? Thanks so much for everything!

kjg:
@jeremy Any guidance on how you'd like me to proceed here?

jeremy:
Merged @ b79fcd2380e92bb830f2c0446e8938cc73dd69c0 updated for #1002.

PR #1001, opened 2016-05-12
Allow the body.raw_source field to be mutated

edwinwills:
With frozen string literals the raw_source field on body can’t be mutated as it’s initialized with an empty frozen string.

This is needed for the Mailkick gem where it attempts to mutate the raw_source for all parts of the email to replace a unsubscribe string with a live token - https://github.com/ankane/mailkick/blob/master/lib/mailkick/processor.rb#L23


jeremy:
Mutating `raw_source` directly is going to be (very) error-prone and generally undesirable. Consider gsubbing against each part's decoded body and assigning the body back to the part. That guarantees that any other memoized vars won't go stale (unaware of the hidden mutation) and will work with other transfer encodings, like base64, that aren't so amenable to direct modification.

PR #1002, opened 2016-05-16
Use the unicode replacement character

kjg:
instead of an empty string so that it is clear that there was some sort of
encoding issue as opposed silently dropping data

(as discussed in #978)

PR #1003, opened 2016-05-16
Fix issue decoding base64 on some rubies

kjg:


kjg:
@jeremy this fixes most of the failures in #1002, mind taking a look?

jeremy:
Nice. Let's match Ruby's unpadded-input behavior: https://github.com/ruby/ruby/commit/6b6680945ed3274cddbc34fdfd410d74081a3e94#diff-af7a52a71eb7592e89c80506e3383d75R97

Looks like we can omit the `base64_needs_padding?` check, too, since padding step will be naturally skipped for an already-padded string.

kjg:
Thanks! That's perfect, I'll do that

kjg:
@jeremy do you happen to know when/where unpack was modified to handle no padding?

jeremy:
@kjg Looked like the commit linked above. Didn't check Ruby commit history exhaustively, though.

kjg:
This is where unpack itself got fixed: https://github.com/ruby/ruby/commit/3e2fdaf1dd76dc39177dce7937897f4dc70e875d

kjg:
> Looks like we can omit the base64_needs_padding? check, too, since padding step will be naturally skipped for an already-padded string.

@jeremy Due to https://github.com/ruby/ruby/commit/3e2fdaf1dd76dc39177dce7937897f4dc70e875d, starting with ruby 2.1 `unpack('m')` handles unpadded characters. That change you linked to in the base64 stdlib is basically unrelated and was only added for ruby 2.3, so without the base64_needs_padding? check, we are adding padding for `unpack('m')` even when we don't *need* to. That method of adding the padding is still better than mine, so I'm definitely going to use that way now, thats for that. I also still think it is probably nicer to not have that check for the version of unpack('m') that doesn't handle no padding, and just always add the padding even when `unpack('m')` doesn't need it.

I've made that change, so if you agree, this should be good to go. If you want me to go back to `base64_needs_padding?` let me know and I'll do that.

jeremy:
Backported 1b420dcb7f94055a143088879e13b43710e5a36b to 2-6-stable too.

Issue #577, opened 2013-07-03
POP3 Retriever Should Not Force SSL Verify None


carsonreinke:
Currently, the POP3 retriever does this whenever SSL is enabled:

``` ruby
pop3.enable_ssl(OpenSSL::SSL::VERIFY_NONE) if settings[:enable_ssl]
```

(https://github.com/mikel/mail/blob/master/lib/mail/network/retriever_methods/pop3.rb#L129)

Should this not be like SMTP delivery method which allows for `:openssl_verify_mode` setting?

jeremy:
Yes it should be! And VERIFY_NONE should certainly not be the default.

PR #1008, opened 2016-06-07
Do not suggest nonsensical OpenSSL verify modes

jhass:
`SSL_set_verify(3)` explains:

```
SSL_VERIFY_FAIL_IF_NO_PEER_CERT
  Server mode: if the client did not return a certificate, the TLS/SSL handshake is immediately terminated with a "handshake failure" alert.  This flag must
  be used together with SSL_VERIFY_PEER.

  Client mode: ignored

SSL_VERIFY_CLIENT_ONCE
  Server mode: only request a client certificate on the initial TLS/SSL handshake. Do not ask for a client certificate again in case of a renegotiation.
  This flag must be used together with SSL_VERIFY_PEER.

  Client mode: ignored
```

The SMTP connection here uses an OpenSSL socket in client mode, suggesting invalid flags is rather misleading.

### DIFF
--- a/lib/mail/network/delivery_methods/smtp.rb
+++ b/lib/mail/network/delivery_methods/smtp.rb
@@ -45,9 +45,8 @@ module Mail
   # hostname or update the certificate authorities trusted by your ruby. If
   # that isn't possible, you can control this behavior with
   # an :openssl_verify_mode setting. Its value may be either an OpenSSL
-  # verify mode constant (OpenSSL::SSL::VERIFY_NONE), or a string containing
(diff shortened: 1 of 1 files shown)


jhass:
Any issue with this? Rails accepted the same change to their documentation.

carsonreinke:
Related https://github.com/mikel/mail/issues/577

jeremy:
Thanks @jhass! Merged @ 120f0b48c32fdab0fe659ff5c2fc15d96a2309ad

PR #1009, opened 2016-06-22
Address list parser should handle /\. +/ in local part of address (WIP)

lukebooth:
I couldn't come up with a solution to this problem, but I've made a PR with failing tests to show what I've found.

It looks like the gem treats an address with 1+ spaces after a `.` in the local side of an address as valid but doesn't return the correct address. Some examples:

```
Mail::Address.new("jlucas. booth@gmail.com") # => One space after a period
=> #<Mail::Address:2236832580 Address: |@gmail.com| >
Mail::Address.new("jlucas.  booth@gmail.com") # => Two+ spaces after a period
=> #<Mail::Address:2236815680 Address: |jlucas.| >
```

I couldn't get `rake ragel:generate` to work in order to make changes and see if the new parser works, so can I get some guidance on fixing this?

### DIFF
--- a/spec/mail/elements/address_spec.rb
+++ b/spec/mail/elements/address_spec.rb
@@ -588,6 +588,32 @@ describe Mail::Address do
                                          :raw          => 'jdoe@test.example'})
       end
 
+      it "should handle |jlucas. booth@test.example|" do
+        address = Mail::Address.new('jlucas. booth@test.example')
+        expect(address).to break_down_to({
+                                         :name         => 'jlucas.booth@test.example',
+                                         :display_name => 'jlucas.booth@test.example',
+                                         :address      => 'jlucas.booth@test.example',
+                                         :comments     => nil,
+                                         :domain       => 'test.example',
+                                         :local        => 'jlucas.booth',
+                                         :format       => 'jlucas.booth@test.example',
+                                         :raw          => 'jlucas. booth@test.example'})
+      end
+
+      it "should handle |jlucas.  booth@test.example|" do
+        address = Mail::Address.new('jlucas. booth@test.example')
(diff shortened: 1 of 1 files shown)


lukebooth:
Could I get some feedback on this PR — what I could do to move it forward, is it an issue the gem should concern itself with, etc — otherwise I can close it as a #willnotfix

jeremy:
Took a look at supporting this in the parser. Tricky because the spaces make it look like part of the display name, but only until an address without angle brackets has been parsed.

Do you have an email in the wild with addresses like these?

lukebooth:
I don't have any examples of this in the wild, unfortunately.
We don't do validations on emails for our apps because they aren't _necessarily_ going to be used for contacting people, sometimes it's just for filing and we decided to just leave those fields as free text.

We've had a few instances where customers fat-finger a space or two in the email after a `.` and when that email _does_ get used for contacting a person, it's kind of a silent fail because the gem doesn't see it as invalid and so no InvalidErrors are raised for us to rescue from.

It's kind of a difficult problem, because _technically_ that is a valid email address — 1-2 spaces after a `.`. But if it's ever used in the real world, I'm not sure. So should we allow it to be valid, get the gem to parse it correctly, and let users get the "your email was undeliverable" response back **or** do we tip our hats to its technical syntactic validity, but we don't support it and an invalid address exception is raised when parsing it?

jeremy:
> technically that is a valid email address — 1-2 spaces after a .. 

By what reading of the RFCs? As far as I can tell, the local part has never allowed spaces.

I think the pragmatic route here is to treat whitespace stripping as a matter of input validation. Clean up common typo issues in user input before parsing.

Thanks @lukebooth!

PR #1011, opened 2016-07-06
Allow reading body_encoding without passing nil

fphilipe:
I assume that a default value of `nil` was forgotten when this method was added, since I expect `message.body_encoding` to return the desired value similar to `message.body`.

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -1247,7 +1247,7 @@ module Mail
       end
     end
 
-    def body_encoding(value)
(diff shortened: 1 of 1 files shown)


jeremy:
I'm puzzled that we have these methods. They're unused by other code, they're untested, and they just proxy to `body.encoding`. Are you putting them to use?

fphilipe:
I was using them just because they are there. I wouldn't mind going through `body.encoding`.

PR #1020, opened 2016-08-18
Don't set OpenSSL verify mode if none is explicitly passed

jhass:
While Net::SMTP.default_ssl_context doesn't return anywhere near sane defaults still,
if somebody was wise enough to monkey patch sane defaults into it, or if it ever changes
to have sane defaults, this shouldn't destroy those efforts by setting verify_mode back to
nil unless the gem is explicitly configured not to.


jeremy:
Sounds good. Test coverage?

jhass:
Updated.

Issue #1031, opened 2016-09-21
If address header cannot be parsed, Mail.new leaves it as a string


cg2v:
I am using mail to parse some incoming messages.
Some of the messages have incorrect To headers and mail's behaviour is strange; it does not split the address list up at all, the to field returns the raw string value of the header:

2.2.0 :027 > Mail.new("To: user1@example.org, user2@example.org").to
 => ["user1@example.org", "user2@example.org"] 
2.2.0 :028 > Mail.new("To: user1@example.org, user2@example.org.").to
 => "user1@example.org, user2@example.org." 

(the error is the trailing dot in example.org; Mail::AddressList.new() cannot parse this string)

keysen:
+1 same thing for me

And it's very annoying because it occurs even when it's well formatted. Here I reproduce a case where it don't work because it's quoted printable with encoding...

For example: 
```ruby
[1] pry(main)> m = Mail.new("To: =?UTF-8?Q?Alain_DUBOUILLON_<contact@dubouillon.fr>,_TLM_<tlm@tlm.fr>,?=\n =?UTF-8?Q?_=222=C2=B0_BUREAU=22_<culture1@2e-bureau.com>,_491_<mensuel491@o?=\n =?UTF-8?Q?range.fr>,_=22ACTUART_=28Eliza_PLOIA=29=22_<elizaploia@yahoo.fr>?=")
[2] pry(main)> m.to
=> "Alain DUBOUILLON <contact@dubouillon.fr>, TLM <tlm@tlm.fr>, \"2° BUREAU\" <culture1@2e-bureau.com>, 491 <mensuel491@orange.fr>, \"ACTUART (Eliza PLOIA)\" <elizaploia@yahoo.fr>"

[3] pry(main)> m = Mail.new("To: Alain DUBOUILLON <contact@dubouillon.fr>, TLM <tlm@tlm.fr>, \"2° BUREAU\" <culture1@2e-bureau.com>, 491 <mensuel491@orange.fr>, \"ACTUART (Eliza PLOIA)\" <elizaploia@yahoo.fr>")
[4] pry(main)> m.to
=> ["contact@dubouillon.fr", "tlm@tlm.fr", "culture1@2e-bureau.com", "mensuel491@orange.fr", "elizaploia@yahoo.fr"]
```

kjg:
So the question is, what SHOULD the behavior be when Mail cannot parse the header. Should it still return an array, but with just one item and that one item is the entire header contents?

[...]

PR #1032, opened 2016-09-23
Re-raise errors in the Message object, fixup message_id inspection

bobjflong:
Fixes https://github.com/mikel/mail/issues/1031

Pulling a bad attribute from a message now re-raises the exception.

``` ruby
Mail::Message.new(bad_email).to
```

This surfaced a bug with message_ids. In https://github.com/bobjflong/mail/blob/ba70362b27a4bcbd020bb632b4e1e2461f29cbcd/lib/mail/message.rb#L368 the original message ids are reset after a comparison. However this reset was using the parsed form instead of the actual header value. This surfaced in specs such as https://github.com/bobjflong/mail/blob/ba70362b27a4bcbd020bb632b4e1e2461f29cbcd/spec/mail/message_spec.rb#L1580.

eg. before comparison:

| Headers |  |
| --- | --- |
| message_id | `<foo@bar.com>` |

after comparison:

| Headers |  |
| --- | --- |
| message_id | `foo@bar.com` |

``` ruby
> m1 = Mail.new("To: mikel@test.lindsaar.net\r\nSubject: Yo!\r\n\r\nHello there")
> m2 = Mail.new("To: mikel@test.lindsaar.net\r\nMessage-ID: <1234@test.lindsaar.net>\r\nSubject: Yo!\r\n\r\nHello there")

> message_id = m2.header[:message_id].value
"<1234@test.lindsaar.net>"
> m1 == m2
true
> message_id == m2.header[:message_id].value
false
```

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -356,11 +356,11 @@ module Mail
     #  m1 == m2 #=> false
     def ==(other)
       return false unless other.respond_to?(:encoded)
-
       if self.message_id && other.message_id
         self.encoded == other.encoded
       else
-        self_message_id, other_message_id = self.message_id, other.message_id
+        self_message_id = header[:message_id] ? header[:message_id].value : nil
+        other_message_id = other.header[:message_id] ? other.header[:message_id].value : nil
         begin
           self.message_id, other.message_id = '<temp@test>', '<temp@test>'
           self.encoded == other.encoded
@@ -1194,7 +1194,7 @@ module Mail
       if val
         header[sym] = val
       else
-        header[sym].default if header[sym]
+        check_errors_and_return_default(header[sym])
       end
     end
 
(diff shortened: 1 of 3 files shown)


lib/mail/message.rb
```diff
@@ -356,11 +356,11 @@ def <=>(other)
     #  m1 == m2 #=> false
     def ==(other)
       return false unless other.respond_to?(:encoded)
-
       if self.message_id && other.message_id
         self.encoded == other.encoded
       else
-        self_message_id, other_message_id = self.message_id, other.message_id
+        self_message_id = header[:message_id] ? header[:message_id].value : nil
```
bobjflong: Given a message id such as `<foo@bar.com>`, the existing code _was_ mutating the object by setting the parsed version directly back into the header. This pulls the value directly from the headers so that they go back the same.

This was surfaced in this bug fix as pulling attributes from the Message class no longer silently fallback to their raw form.

spec/mail/message_spec.rb
```diff
@@ -1987,7 +1992,7 @@ def self.delivering_email(mail)
       before do
         @mail = Mail.new do
           in_reply_to '<1234@test.lindsaar.net>'
-          message_id '5678@test.lindsaar.net'
+          message_id '<5678@test.lindsaar.net>'
```
bobjflong: Fix fixture needed to be fixed.

jeremy:
Raising on all parse errors is at odds with the Mail lib design, and a major breaking change.

We can fix the lossy Message ID after comparison bug by duping the message before comparing instead of using its headers as scratch space.

PR #1034, opened 2016-09-28
Allowed Mail::Address.new to accept non US ASCII strings

ben5516:
Currently, if you instantiate an address list with a non US ASCII character you get a parse error. This PR fixes that problem by calling `Mail::Encodings.encode_non_usascii` _before_ Ragel tries to parse the address.

### DIFF
--- a/lib/mail/parsers/address_lists_parser.rb
+++ b/lib/mail/parsers/address_lists_parser.rb
@@ -14051,6 +14051,7 @@ self.en_main = 1334;
     def self.parse(data)
       address_list = AddressListStruct.new([], [])
       return address_list if Mail::Utilities.blank?(data)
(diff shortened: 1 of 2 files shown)


jeremy:
Thanks @ben5516. Downgrading i18n headers before parsing is a good approach, barring UTF8 parsing. Check out https://tools.ietf.org/html/rfc5504 for a full rundown—there's a good bit more to cover.

shved:
@jeremy Hi! Could you please tell me what's up with this PR? Cant find it in 2.7.0 PR list. Waiting for such update

jeremy:
@shved Note that this PR is open and hasn't been merged. There's a partial implementation that'll need more work to complete.

jeremy:
Fixed by #1103 🖖

Issue #1035, opened 2016-09-30
Reading IMAP sent emails


dvisockas:
Is there currently a way to read sent emails?

In _lib/mail/network/retriever_methods/imap.rb_ there is a line `options[:mailbox] ||= 'INBOX'` but changing it to either 'sent' or 'SENT' changes nothing.

jeremy:
What's the name of your "Sent Messages" folder? Is it actually "SENT?"

Try this out with other folders. The Sent folder isn't special; it's a convention for where to keep sent emails and each mail app does it differently. Many allow you to map a folder of your choice for use as Sent, Trash, Junk, etc.

dvisockas:
It's actually SENT.

Then let me rephrase my question - is there a way to get the folder names in a mailbox?

jeremy:
@dvisockas Yep! The Mail lib uses [Ruby's Net::IMAP](http://ruby-doc.org/stdlib-2.3.1/libdoc/net/imap/rdoc/Net/IMAP.html#method-i-list), so you have access to all its methods:
```ruby
# Retrieve the top-level folders by matching with % instead of *
mailboxes = Mail.connection { |imap| imap.list '', '%' }
```
IMAP servers that support [RFC6154](https://tools.ietf.org/html/rfc6154) will label their Sent folder with a `\Sent` flag which we can check for:
```ruby
sent = mailboxes.detect { |m| m.attr.include? :Sent }
```
Other IMAP servers may support the older `XLIST` command which does the same thing:
```ruby
sent = Mail.connection { |imap| imap.xlist '', '%' }.detect { |m| m.attr.include? :Sent }
```

PR #1037, opened 2016-10-03
Get a list of mailboxes, if retriever supports them

simonasdev:
#1035

### DIFF
--- a/lib/mail/mail.rb
+++ b/lib/mail/mail.rb
@@ -3,45 +3,45 @@
 module Mail
 
   # Allows you to create a new Mail::Message object.
-  # 
+  #
   # You can make an email via passing a string or passing a block.
-  # 
+  #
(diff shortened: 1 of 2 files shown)


jeremy:
Does this make sense for other retriever methods, or only IMAP? I like the convenience but it feels like a leaky abstraction since you need to know to work with Net::IMAP to make sense of the return values.

simonasdev:
You might be right, this is mostly for the convenience so I decided to put it here. We could return nil or even raise something like 'Protocol doesn't support mailboxes', if you think it'll make more sense

jeremy:
Dig the convenience, but it broadens our toplevel Mail API without the kind of underlying support we'd expect: that it'd work for all retriever methods. I think dropping down to IMAP is appropriate in this case, as in #1035. Thanks @simonasdev!

PR #1042, opened 2016-10-17
Added RDoc for Mail::Parsers::ContentTransferEncodingParser.parse

vivganes:


jeremy:
Thanks for the patch! However, this is an internal class, not meant to be used directly, and its purpose is pretty clear without rdoc 😊

PR #1047, opened 2016-11-11
Fixing #408 by not touching the body of the email

jniebuhr:
The mail parsing was rewritten to use streams in order to parse headers line by line. After parsing the headers, the rest of the stream will be just read into the body.


jniebuhr:
1.8.7 and 1.9.2 tests seem to fail because of the rdoc dependency.

jniebuhr:
Did somebody take a look at this yet?

spec/fixtures/emails/rfc2822/example14.eml
```diff
@@ -3,9 +3,8 @@ Date: Fri, 19 Aug 2011 10:47:17 +0900
 From: Atsushi Yoshida <atsushi@example.com>
 Reply-To: rudeboyjet@gmail.com
 Subject: Re: TEST
- 
-	=?ISO-2022-JP?B?GyRCJUYlOSVIGyhC?=
-  =?ISO-2022-JP?B?GyRCJUYlOSVIGyhC?=
+ =?ISO-2022-JP?B?GyRCJUYlOSVIGyhC?=
+ =?ISO-2022-JP?B?GyRCJUYlOSVIGyhC?=
```
jeremy: Why'd you need to change this test fixture?

spec/mail/attachments_list_spec.rb
```diff
@@ -306,6 +306,13 @@ def check_decoded(actual, expected)
       expect(mail.attachments[0].filename).to eq "Foto0009.jpg"
     end
 
+    it 'should be able to read binary parts (issue 408)' do
+      mail = Mail.read(fixture(File.join("emails",
+                                         "attachment_emails",
+                                         "attachment_pdf_binary.eml")))
+      expect(mail.attachments.length).to eq 1
+      expect(mail.attachments[0].body.raw_source.length).to eq 11464
```
jeremy: Should this use `#bytesize`?

lib/mail/message.rb
```diff
@@ -1994,14 +1992,44 @@ def text?
...
+    end
+
+    def headers_required?
+      true
     end
 
     def raw_source=(value)
       value = value.dup.force_encoding(Encoding::BINARY) if RUBY_VERSION >= "1.9.1"
-      @raw_source = ::Mail::Utilities.to_crlf(value)
+      @raw_source = value
```
jeremy: ❤️

lib/mail/message.rb
```diff
@@ -1994,14 +1992,44 @@ def text?
...
+          line = line.strip if first
+          buffer += line
+        else
+          buffer += CRLF unless first
+          buffer += line
+        end
+        first = false
+      end
+      buffer
+    end
```
jeremy: This is difficult to check for correctness. Please comment the the implementation to indicate intent and logic flow.

lib/mail/message.rb
```diff
@@ -1994,14 +1992,44 @@ def text?
...
-      header_part, body_part = raw_source.lstrip.split(HEADER_SEPARATOR, 2)
-      self.header = header_part
-      self.body   = body_part
+      stream = StringIO.new(raw_source)
+      self.header = parse_headers(stream)
+      self.body = stream.read
+    end
+
+    # Parses the headers from the input stream,
+    # leaving the position at the start of the body.
```
jeremy: Could you explain your motivation for consuming a stream rather than scanning with a regexp?

lib/mail/part.rb
```diff
@@ -102,16 +102,15 @@ def get_return_values(key)
...
-      header_part, body_part = raw_source.split(/#{CRLF}#{WSP}*#{CRLF}/m, 2)
-      if header_part =~ HEADER_LINE
-        self.header = header_part
-        self.body   = body_part
-      else
-        self.header = "Content-Type: text/plain\r\n"
-        self.body   = raw_source
-      end
+      super
+      self.header['Content-Type'] = 'text/plain' if self.header['Content-Type'].nil?
```
jeremy: Equivalent to?
```ruby
header['Content-Type'] ||= 'text/plain'
```

spec/fixtures/emails/rfc2822/example14.eml
jniebuhr: The empty line doesn't seem to be valid in a header. So I removed that

lib/mail/message.rb
jniebuhr: I figured it would be the best way to actually leave the body untouched since I'm only taking the stuff away I don't need from the stream. Splits and replaces might lead ruby to do weird stuff with the binary content.

lib/mail/message.rb
jniebuhr: I've commented it in detail now

spec/fixtures/emails/rfc2822/example14.eml
jeremy: That change appears to be unrelated to this PR, though. Would limit scope to just the change at hand.

lib/mail/message.rb
jeremy: How would it lead Ruby to do weird stuff?

lib/mail/message.rb
jeremy: Thanks!

spec/fixtures/emails/rfc2822/example14.eml
jniebuhr: my changed version of the algorithm does not parse this anymore

lib/mail/message.rb
jniebuhr: i've found it sometimes strips bytes when detecting encoding of a string

PR #1048, opened 2016-11-13
Fix documentation on Address.comments methods [skip ci]

esparta:
Currently the documentation on the method `comments` says it will return an empty array
if there is not comments on `Mail::Address`. This is not accurate on current implementation
and confirmed also on current test ( mail/spec/mail/elements/address_spec.rb#213 - "email addresses from the wild") where we expect `nil` if the address has no comments


jeremy:
Thanks @esparta! Merged @ 19597dcc182b5cf2942887bdf6cf5e18da4aeebc

esparta:
Thanks for your time reviewing it and merging.

PR #1051, opened 2016-11-23
Handle botched Content-Type ending in \r\n

aotenko:
Problem: Seen mail with Content-Type: text/html Content-Transfer-Encoding: 8bits\r\n (with \r\n spelled literally like that)

Solution: sanatize() to take care of this case

### DIFF
--- a/lib/mail/fields/content_type_field.rb
+++ b/lib/mail/fields/content_type_field.rb
@@ -144,10 +144,9 @@ module Mail
 
       # TODO: check if there are cases where whitespace is not a separator
       val = val.
+        gsub(/\s[\w\-]+:\s.*\\r\\n$/, ''). # cases with botched fields like Content-Type: text/html Content-Transfer-Encoding: 8bit\r\n
         gsub(/\s*=\s*/,'='). # remove whitespaces around equal sign
-        tr(' ',';').
-        squeeze(';').
-        gsub(';', '; '). #use '; ' as a separator (or EOL)
+        gsub(/[; ]+/, '; '). #use '; ' as a separator (or EOL)
         gsub(/;\s*$/,'') #remove trailing to keep examples below
 
       if val =~ /(boundary=(\S*))/i
@@ -157,9 +156,6 @@ module Mail
       end
 
       case
-      when val.chomp =~ /^\s*([\w\-]+)\/([\w\-]+)\s*;;+(.*)$/i
-        # Handles 'text/plain;; format="flowed"' (double semi colon)
-        "#{$1}/#{$2}; #{$3}"
       when val.chomp =~ /^\s*([\w\-]+)\/([\w\-]+)\s*;\s?(ISO[\w\-]+)$/i
         # Microsoft helper:
         # Handles 'type/subtype;ISO-8559-1'
@@ -186,7 +182,7 @@ module Mail
       when val =~ /^\s*$/
         'text/plain'
       else
-        ''
+        val
       end
     end
 
--- a/spec/fixtures/emails/error_emails/content_type_crlf.eml
+++ b/spec/fixtures/emails/error_emails/content_type_crlf.eml
@@ -1,0 +1,101 @@
+From root@smtp.nntc.net  Thu Feb  2 01:05:21 2006
+Return-Path: <root@smtp.nntc.net>
+X-Original-To: nobody@login.example.com
+Delivered-To: nobody@login.example.com
+Received: from mail2.example.com (mail2.example.com [10.3.4.5])
+	by naughty.example.com (Postfix) with ESMTP id 44F0E5371FD
+	for <nobody@login.example.com>; Thu,  2 Feb 2006 01:05:21 -0500 (EST)
+Received: from smtp.nntc.net (12.186.41.2.nntc.net [12.186.41.2])
+	by mail2.example.com (Postfix) with ESMTP id 080E06FA302
+	for <nobody@example.com>; Thu,  2 Feb 2006 01:05:21 -0500 (EST)
+Received: from smtp.nntc.net (rebuilt.nntc.net [127.0.0.1])
+	by smtp.nntc.net (8.12.11/8.12.8) with ESMTP id k1265FsI017665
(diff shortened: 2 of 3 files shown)


lib/mail/fields/content_type_field.rb
```diff
@@ -144,10 +144,9 @@ def sanatize( val )
 
       # TODO: check if there are cases where whitespace is not a separator
       val = val.
+        gsub(/\s[\w\-]+:\s.*\\r\\n$/, ''). # cases with botched fields like Content-Type: text/html Content-Transfer-Encoding: 8bit\r\n
         gsub(/\s*=\s*/,'='). # remove whitespaces around equal sign
-        tr(' ',';').
```
aotenko: tr all spaces to ;, then replace all sequences of ; with a single '; ' is the same as gsub(/[; ]+/, '; ')

aotenko:
"All checks have failed" for some generic reason not related to the change - build failed to install stuff, or decode some Chinese.

lib/mail/fields/content_type_field.rb
```diff
@@ -144,10 +144,9 @@ def sanatize( val )
 
       # TODO: check if there are cases where whitespace is not a separator
       val = val.
+        gsub(/\s[\w\-]+:\s.*\\r\\n$/, ''). # cases with botched fields like Content-Type: text/html Content-Transfer-Encoding: 8bit\r\n
```
jeremy: This is a very specific edge case for a very broken mail header. Do you think we should support it?

lib/mail/fields/content_type_field.rb
jeremy: 👍

lib/mail/fields/content_type_field.rb
```diff
@@ -157,9 +156,6 @@ def sanatize( val )
       end
 
       case
-      when val.chomp =~ /^\s*([\w\-]+)\/([\w\-]+)\s*;;+(.*)$/i
```
jeremy: 👍

lib/mail/fields/content_type_field.rb
aotenko: We are processing all sorts of email. What is a mitigation for cases when we encounter an email with the header like this?

jeremy: We need to strike a balance between "seen in one email we failed to parse" and "things like common issue with emails from Hotmail."

We'd be likely to implement a workaround for common, recurring brokenness, but unlikely to take on the code baggage for rare errors. Particularly when the workaround may cause issues with other valid emails.

In this case, raising an exception and allowing parsing to fail is probably appropriate. Then you can address and fix the problem in your app, likely by correcting the email manually if needed.

We can't promise to parse every email under the sun. But we can make a reasonable best effort :)

aotenko: ...and what exactly will I do, when an Exception of this sort occurs? Not deliver the email?

Some emails are _constructed_ to work around strict parsers. We have several thousand emails in our regression suite. Broken headers and invalid encodings constitute an important part of that suite.

You don't have to agree with our vision of what this library is supposed to handle. We'll just have to live with a fork.

jeremy: Pardon, we haven't met. I don't know how you're using this library or what you need from it, so you're leaving me at a loss with the weight of your concerns and expectations.

Here's some background on the vision of this library, since you're implying that it's a strict parser, which is definitively is not.

The Mail library follows [Postel's classic principle](https://en.wikipedia.org/wiki/Robustness_principle): "Be conservative in what you do, be liberal in what you accept from others." We have a long history of liberal parsing to fulfill our aim to be very user-friendly to developers.

The Mail library should generally parse just about any email in the wild. You can see this demonstrated in our rake tasks that successfully parse a huge corpus of spam email.

This gets to the balance I discuss above. How liberal can we be without compromising valid email parsing (by introducing new bugs!) and taking on undue code complexity by littering our codebase with workarounds for rare, exceptional emails? Broken charsets and encodings are a great example. We go to great lengths to transcode to Ruby string encodings, but what is the right choice when we cannot? Should we make a best guess, force the encoding, raise an exception? These are policy decisions that one user may find as a blessing where another user finds them as a curse.

Our best option is to be reasonable and sensible about striking a balance. We can attempt to accept all the emails we can without making outsized compromises for very rare or exceptional cases. If you fall on the other side of this decisionmaking because you have such an exceptional email and feel the built-in parser should correct it—even with a single-case workaround that threatens new bugs and parsing errors with other valid emails—then it'll always be hard to live with this tradeoff. But: I hope we can agree it's a necessary and healthy tradeoff to make.

Please consider sharing your regression suite (with personal/private info scrubbed). Perhaps we can address other brokenness and help relieve your workaround burden from maintaining a fork.

aotenko: Sure, I do understand you want to strike the right balance for your vision of the library's future.

All I am saying is that throwing an Exception is not an answer to my quest. If an Exception is raised, the caller simply has no option but to fail the processing completely. The caller cannot fixup-and-proceed-from-where-it-failed, because there is no way of knowing where in the email body processing has failed and in what way. So I cannot fix this problem by acrobatics in my code or by fixing up the email body.

I don't see a problem that you won't merge my pull request. I am not trying to persuade you to change your acceptance criteria.

If you see some emails from your regression suite for which the change fails, it'd be great to see those emails, too.

jeremy: > All I am saying is that throwing an Exception is not an answer to my quest. If an Exception is raised, the caller simply has no option but to fail the processing completely. The caller cannot fixup-and-proceed-from-where-it-failed, because there is no way of knowing where in the email body processing has failed and in what way. So I cannot fix this problem by acrobatics in my code or by fixing up the email body.

I'm sympathetic to this and I've encountered the same pattern. I'd often prefer to have an unparsable field marked with an error rather than raise an exception. That defers the issue from when the email is parsed to when the Mail user attempts to access the field, having expected it to be parsed.

Curious, if you are free to share: Are you implementing a mailing list or remailer? Or a mail filter/firewall?

PR #1059, opened 2016-12-14
FEATURE: remove the dependency on the mime types use mini_mime

SamSaffron:
The mime types gem (even with the magical columnar store by @jeremyevans) is a **memory hog**.

In particular on boot, mime/types/columnar will allocate **109k** objects and end up with a retained count of **31K** objects. 

This bloat leaves objects that need to be marked and swept every major GC and bloats ruby processes. 

To circumvent mini_mime was created. 

It uses the exact same database as the full fledged mime types and is capable of only loading stuff on demand with a practical, safe and bound in-memory cache. 

mini_mime will allocate **398** objects on boot and only retain **62** (due to rubygems inefficiency) 

In table form:

| |boot allocations|boot retained|lookup|lookup uncached|
|-|-:|-:|-:|-:|
|mini_mime|398|62|641K/s|33K/s|
|mime-types|109796|31165|361K/s|361K/s|

The performance of mini_mime is pretty good, cached lookups are faster than the mime types gem and uncached lookups are 10x slower. (which is really not at huge issue considering the memory savings)


mperham:
my hero

SamSaffron:
FYI did a similar PR to rest-client

https://github.com/rest-client/rest-client/pull/557

radar:
This is great! Thank you @samsaffron. 

Is there any particular reason why you're removing the older versions of Ruby from the travis config? Seems out of scope for this PR.

SamSaffron:
@radar it was a bit fiddly to remove the giant matrix of mime types tests and this seemed like the simplest way to resolve, when I removed stuff and left the rake hack in the Gemfile it had issue running specs

supporting 1.8 or 1.9 seems very very bad these days ... 1.9 has no support or security patches for almost 2 years.

ultimately it is your call if to support 1.8 but would appreciate some help figuring out an appropriate Gemfile hack here.

jeremy:
Squashed with #1060 and merged @ 634d91145cb65a85dff9da61eddc9e2de0609f8b 🖖

czj:
Thanks a lot. Ruby and Rails need this kind of optimisation !

schneems:
Sam you are the `mini_*` master 😎

stereobooster:
Am I missing something or it is not released yet? https://rubygems.org/gems/mail/versions/2.6.4

```
VERSIONS:
2.6.4 - March 23, 2016 (344 KB)

RUNTIME DEPENDENCIES (1):
mime-types < 4, >= 1.16
```

PR #1060, opened 2016-12-14
mini_mime without dropping legacy Ruby support

jeremy:
#1059


jeremy:
In tandem with https://github.com/discourse/mini_mime/pull/1

jeremy:
Merged @ 634d91145cb65a85dff9da61eddc9e2de0609f8b

stereobooster:
What prevents it get to 2.7.0 release? I mean there is only RC (since may), but no final version of gem.

I have implemented [micro_mime](https://github.com/stereobooster/micro_mime), which is the same as `mini_mime`, but consumes even less memory and faster. The question is how to trick `mail` gem to use `micro_mime` instead of `mini_mime`? Something like `multi_json`, but for mime gems

jeremy:
@stereobooster, track the 2.7.0 milestone at https://github.com/mikel/mail/milestone/5

re. mini_mime vs micro_mime, [my comment on the earlier pull request](https://github.com/discourse/mini_mime/pull/5#pullrequestreview-32675071): 
> I think most users would prefer to have an extension bundled with the library itself rather than needing to know to add an additional optional dependency. That'll also tempt other library authors to depend on both mini_mime and micro_mime with awkward/mismatched version constraints.

This is exactly the situation I—and other library authors, I wager!—would like to avoid. We dropped mime-types for mini_mime rather than introduce some mime lookup plugin interface, after all.

ioquatix:
My opinion is that mail depending on ANY mime type library is an anti-pattern. Make it a lambda function you provide as part of the configuration, then the user can supply whatever code they prefer to do mime-type lookup.

stereobooster:
How about "mail-core" gem, which will not have this dependency and mail gem which will depend on  `mini_mime` and "mail-core"

jeremy:
@stereobooster The simplest route is to have users `require 'micro_mime'` after requiring mini_mime. Then you can replace the MiniMime constant with your MicroMime implementation.

PR #1061, opened 2016-12-14
fix invalid encoding

ahorek:
error: ArgumentError (invalid byte sequence in UTF-8)

'test'.force_encoding('windows-1258').encode('UTF-8') -> Encoding conversion failed code converter not found (Windows-1258 to UTF-8)

similar issue for utf-7: https://github.com/mikel/mail/pull/650
some character may be lost this way, but I think it's better than a decoded output with invalid encoding. Another way is to map windows-1258 to windows-1252? What do you think?

### DIFF
--- a/lib/mail/version_specific/ruby_1_9.rb
+++ b/lib/mail/version_specific/ruby_1_9.rb
@@ -97,7 +97,8 @@ module Mail
       decoded.valid_encoding? ? decoded : decoded.encode(Encoding::UTF_16LE, :invalid => :replace, :replace => "").encode(Encoding::UTF_8)
     rescue Encoding::UndefinedConversionError, ArgumentError, Encoding::ConverterNotFoundError
       warn "Encoding conversion failed #{$!}"
-      str.dup.force_encoding(Encoding::UTF_8)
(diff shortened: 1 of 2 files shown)


jeremy:
Agree, but we're stuck with API compatibility for a good while. Lossy transcoding won't fly.

Check out #785 to provide your own charset encoder. BestEffortCharsetEncoder maps Windows-1258 to Windows-1252, for example.

ahorek:
thanks @jeremy , but it doesn't work now even if BestEffortCharsetEncoder is set manually.
As you can see on this line, charset_encoder isn't used.

decoded = str.encode(Encoding::UTF_8, :undef => :replace, :invalid => :replace, :replace => "")

woodie:
I made local delivery work on windows like this...
https://gist.github.com/woodie/363e76e589aa2b642ad320dab9b4682d

jeremy:
@ahorek Note that's in the `rescue` clause after calling `charset_encoder.encode(…)`, however. By providing your own charset encoder, you can ensure you don't raise one of those exceptions so you don't hit the "force-to-utf8" fallback.

PR #1064, opened 2017-01-01
CI against Ruby 2.4.0

koic:
Ruby 2.4.0 has been released. Then this Ruby version is [available on Travis CI](http://rubies.travis-ci.org).

https://www.ruby-lang.org/en/news/2016/12/25/ruby-2-4-0-released

Thanks.


koic:
Thanks for the merging!

PR #1065, opened 2017-01-01
Implement enforcing STARTTLS mode.

### LINKED ISSUE #521: Support an "always" mode for STARTTLS.
Feature Suggestion: Support an "always" mode for STARTTLS in the SMTP delivery method.

If `:enable_starttls_always` were true, then instead of calling [`smtp.enable_starttls_auto`](https://github.com/mikel/mail/blob/2.4.4/lib/mail/network/delivery_methods/smtp.rb#L138), the `#deliver!` method would call [`smtp.enable_starttls`](https://github.com/ruby/ruby/blob/v1_9_3_392/lib/net/smtp.rb#L334-L341).

This setting will allow the mail sender to be sure that STARTTLS happened and succeeded before sending the AUTH credentials and message content over the wire.

This setting would be used by anybody who is a priori sure that the SMTP server he intends his application to talk to is capable of performing the STARTTLS session protocol upgrade.

This prevents a hypothetical MITM attack where the attacker strips out the advertised STARTTLS support from the server's initial advertisements list to trick the client into not trying to perform a STARTTLS (because the server sent out the STARTTLS advertisement, the attacker stripped it out, and the client never got it).

### LINKED ISSUE #694: Security of enable_starttls_auto vs. enable_starttls
Looking at https://github.com/mikel/mail/blob/master/lib/mail/network/delivery_methods/smtp.rb#L100-109 it seems that only Net::SMTP's `.enable_starttls_auto` is supported, but not `.enable_starttls`.

As I understand it the auto-method is susceptible to downgrade attacks, thus voiding all advantages of authentication and only leaving passive security in place.

bk2204:
Add an option to always require STARTTLS over SMTP.  This prevents
sending of mail over an unencrypted connection if a third party has
obscured the STARTTLS header (as some firewalls do).

Fixes #521 and fixes #694.

### DIFF
--- a/lib/mail/network/delivery_methods/smtp.rb
+++ b/lib/mail/network/delivery_methods/smtp.rb
@@ -83,6 +83,7 @@ module Mail
                         :user_name            => nil,
                         :password             => nil,
                         :authentication       => nil,
+                        :enable_starttls      => nil,
                         :enable_starttls_auto => true,
                         :openssl_verify_mode  => nil,
                         :ssl                  => nil,
@@ -101,6 +102,10 @@ module Mail
       if settings[:tls] || settings[:ssl]
         if smtp.respond_to?(:enable_tls)
           smtp.enable_tls(ssl_context)
+        end
+      elsif settings[:enable_starttls]
+        if smtp.respond_to?(:enable_starttls)
+          smtp.enable_starttls(ssl_context)
         end
       elsif settings[:enable_starttls_auto]
         if smtp.respond_to?(:enable_starttls_auto)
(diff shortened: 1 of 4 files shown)


jeremy:
Thanks @bk2204. Could you add tests to cover this option?

bk2204:
Sure.  It may take a day or two for me to get back around to this, but I'll add the tests and fix the existing failures.  I'm unclear as to why they didn't fail on my box when I first ran rspec, but I'll check into that, too.

bk2204:
I've improved the tests to perform some basic checking of TLS options for SMTP as a preparatory commit, and then added tests for this option.

jeremy:
Merged @ 646a234ff1278d26c4b4b3fee1e336c59b3c375e

PR #1069, opened 2017-01-09
Port RSpec should usage to expect.

bk2204:
A few specs were still using the older RSpec should syntax, which caused deprecation warnings.  Port these specs to use the expect syntax instead.


jeremy:
Merged @ 3205efa1c20e76a1f7562c913b6182137ba5e8b5 and backported to 2-6-stable @ e76937c24b4c794c832dbdb7636be00c70497d7e

PR #1074, opened 2017-01-26
Fix encoding multiple addresses when noninitial address contains non-usascii chars

domininik:
Previous logic of `Encodings.encode_non_usascii` method used the incorrect regular expression `/(".*?[^\x00-\x7F].*?")/` to find all strings containing non-usascii chars. It matched everything which was before the non-usascii char, even it was a separate address (!).
This resulted in encoding the whole matched group and losing this address (and as consequence not sending an email to this address at all).

One of the possible solutions would be to change the regexp and exclude the quotation mark at the beginning: `("[^"]*?[^\x00-\x7F].*?")` but the regexp can be even more simplified basing on the fact that `Encodings.b_value_encode` does not encode ascii only strings,
so it does not matter if the string passed to it contains the non-usascii chars or not.
The simplified regexp can just match any string which is between the quotation marks: `/(".*?")/`

### DIFF
--- a/lib/mail/encodings.rb
+++ b/lib/mail/encodings.rb
@@ -178,9 +178,8 @@ module Mail
 
     def Encodings.encode_non_usascii(address, charset)
       return address if address.ascii_only? or charset.nil?
-      us_ascii = %Q{\x00-\x7f}
-      # Encode any non usascii strings embedded inside of quotes
-      address = address.gsub(/(".*?[^#{us_ascii}].*?")/) { |s| Encodings.b_value_encode(unquote(s), charset) }
+      # Encode all strings embedded inside of quotes
+      address = address.gsub(/(".*?")/) { |s| Encodings.b_value_encode(unquote(s), charset) }
       # Then loop through all remaining items and encode as needed
       tokens = address.split(/\s/)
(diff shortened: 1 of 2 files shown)


lib/mail/encodings.rb
```diff
@@ -178,9 +178,8 @@ def Encodings.address_encode(address, charset = 'utf-8')
 
     def Encodings.encode_non_usascii(address, charset)
       return address if address.ascii_only? or charset.nil?
-      us_ascii = %Q{\x00-\x7f}
-      # Encode any non usascii strings embedded inside of quotes
-      address = address.gsub(/(".*?[^#{us_ascii}].*?")/) { |s| Encodings.b_value_encode(unquote(s), charset) }
+      # Encode all strings embedded inside of quotes
+      address = address.gsub(/(".*?")/) { |s| Encodings.b_value_encode(unquote(s), charset) }
```
jeremy: Good case to eliminate backtracking with the lazy `"[^"]*"` match instead of the non-greedy `".*?"`.

jeremy:
Merged @ e95bf67d8c1f25fa96972a98dd658c86c32349e5

PR #1077, opened 2017-02-03
Unicode support for local part of email addresses

tayler1:


jeremy:
Nice work @tayler1, thanks for digging in! Would you take a look at #1103? Some of these specs "just work" now 😊

masylum:
Is there any way we can help to move this forward? I'm currently having problems of spanish customers with `ñ` in their emails not going through.

jeremy:
Working on 2.7.0.rc2. I updated the test cases and merged them in. Thanks again @tayler1!

arnt:
Hi,

the code in this pull request misunderstands the RFCs, I'm afraid.

It encodes localparts, and nothing in RFCs 6530-6533 permits that: There's no text along the lines of "parsers MUST decode localparts using such-and-such algorithm" anywhere in any of those documents.

RFC 2047 also forbids the encoding used unambiguously, see [the top of page 8](https://www.rfc-editor.org/rfc/rfc2047#page-8):
    `An 'encoded-word' MUST NOT appear in any portion of an 'addr-spec'.`

I'll submit a new PR to bring message generation into RFC compliance. It may take a few days.

@masylum can you elaborate on what message weren't deliverable? I realise that it's a long time ago.

PR #1082, opened 2017-03-10
Addition of Contents list and markdown adjustments

Minimalistic:
Added a table of contents with relative linking and reformatted headers of the various sections for consistency.


jeremy:
Thanks @Minimalistic!

PR #1084, opened 2017-03-15
Support starttls connections

dopykuh:
Provide ability to serve "starttls_options", if set ssl will be disabled.

### DIFF
--- a/lib/mail/network/retriever_methods/imap.rb
+++ b/lib/mail/network/retriever_methods/imap.rb
@@ -153,8 +153,13 @@ module Mail
       # Start an IMAP session and ensures that it will be closed in any case.
       def start(config=Mail::Configuration.instance, &block)
         raise ArgumentError.new("Mail::Retrievable#imap_start takes a block") unless block_given?
-
-        imap = Net::IMAP.new(settings[:address], settings[:port], settings[:enable_ssl], nil, false)
+        tmp_ssl = settings[:enable_ssl]
+        tmp_ssl = false unless settings[:starttls_options].nil?
+        
+        imap = Net::IMAP.new(settings[:address], settings[:port], tmp_ssl, nil, false)
+        if settings[:starttls_options].is_a?(Hash)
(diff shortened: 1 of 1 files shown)


lib/mail/network/retriever_methods/imap.rb
```diff
@@ -153,8 +153,13 @@ def validate_options(options)
       # Start an IMAP session and ensures that it will be closed in any case.
       def start(config=Mail::Configuration.instance, &block)
         raise ArgumentError.new("Mail::Retrievable#imap_start takes a block") unless block_given?
-
-        imap = Net::IMAP.new(settings[:address], settings[:port], settings[:enable_ssl], nil, false)
+        tmp_ssl = settings[:enable_ssl]
+        tmp_ssl = false unless settings[:starttls_options].nil?
```
jeremy: Since these options are mutually exclusive, we can raise an ArgumentError if both are set.

lib/mail/network/retriever_methods/imap.rb
```diff
@@ -153,8 +153,13 @@ def validate_options(options)
...
+        
+        imap = Net::IMAP.new(settings[:address], settings[:port], tmp_ssl, nil, false)
+        if settings[:starttls_options].is_a?(Hash)
```
jeremy: Providing an options hash to enable/disable the feature is a bit awkward, particular if you don't need to set any options. Would expose this as a boolean: `starttls: true`

lib/mail/network/retriever_methods/imap.rb
```diff
@@ -153,8 +153,13 @@ def validate_options(options)
...
+          imap.starttls(settings[:starttls_options])
```
jeremy: These options aren't unique to STARTTLS. They're also used when SSL is enabled. Both establish a TLS session on the socket.

That suggests calling this `tls_options: { … }` and allowing it to be used in conjunction with both IMAPS and STARTTLS.

jeremy:
#723

PR #1089, opened 2017-04-15
Add required set_params call for system defaults

nanaya:
References:
- https://ruby-doc.org/stdlib-2.4.0/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html#method-i-set_params
- https://gitlab.com/gitlab-org/gitlab-ce/issues/1558#note_1185132


nanaya:
Unfortunately this will set `verify_mode` to system default as well which is `peer` in my system.

lib/mail/network/delivery_methods/smtp.rb
```diff
@@ -141,11 +141,16 @@ def ssl_context
...
-      context
+      Net::SMTP.default_ssl_context.tap do |context|
+        context.verify_mode = openssl_verify_mode if openssl_verify_mode
+
+        if settings[:ca_path] || settings[:ca_file]
+          context.ca_path = settings[:ca_path] if settings[:ca_path]
+          context.ca_file = settings[:ca_file] if settings[:ca_file]
+        elsif openssl_verify_mode && openssl_verify_mode != OpenSSL::SSL::VERIFY_NONE
+          context.cert_store = OpenSSL::SSL::SSLContext::DEFAULT_CERT_STORE
+        end
```
jeremy: Can we always set the cert store? Does it need to be mutually exclusive with provided CA path/file settings?

nanaya: Good question. Unfortunately I'm not sure myself. I based this off `#set_params` which checks all three settings (including `cert_store` itself).

PR #1090, opened 2017-04-17
Remove trailing new lines in header fields

coderob:
When there are trailing new lines in header fields, the field does not currently get parsed.

i.e.,
```
"To: Mikel\r\nFrom: bob\r\n\r\n     \r\n     Message-ID: 1234"
```
In the above example, `Message-ID` will not get parsed, and the resulting Mail object does not contain a message ID.


coderob:
Closing this, as I learned this is NOT even valid according to the RFC spec.

PR #1093, opened 2017-05-04
Remove core_extensions only relevant for Ruby < 2.0

metcalf:
Several of the core_extensions monkey patches are only relevant for Ruby < 2.0. Ruby 1.9.3 has been unsupported for >2 years so I'd propose removing support and the associated extensions.


metcalf:
I noticed that Ruby 1.x is referenced in several places in the documentation. If you agree with the spirit of this change, I can update those sites in the documentation as part of this PR.

jeremy:
We purposefully support Ruby 1.8.7+. We'll drop support for end-of-life Ruby versions in Mail 3.0 😊

andrew-stripe:
Sounds good, feel free to close and I'll tackle this with the approach you recommended yesterday.

andrew-stripe:
Gah, sorry @andrew-stripe == @metcalf... I can never keep my work and personal accounts straight.

PR #1094, opened 2017-05-04
Remove unused methods in core_extensions

metcalf:
The `at`, `from`, `last` and `is_utf8?` patches to String in core_extensions are not currently used by mail.

`to` is only used by `first` for Ruby < 2.0 but I kept both implementations for consistency. `first` is only used in one location and only by Ruby < 2.0 but I also kept that to keep this PR focused on removed completely unused patches.

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -13,6 +13,7 @@ Performance:
 
 Compatibility:
 * #655 - Sort attachments to the end of the parts list to work around email clients that may mistake a text attachment for the message body. (npickens)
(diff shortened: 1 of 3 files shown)


CHANGELOG.rdoc
```diff
@@ -13,6 +13,7 @@ Performance:
 
 Compatibility:
 * #655 - Sort attachments to the end of the parts list to work around email clients that may mistake a text attachment for the message body. (npickens)
+* Removed `at`, `from`, `last` and `is_utf8?` methods from String.
```
jeremy: Users reading this will want to know why they were removed.

metcalf: Updated.

PR #1095, opened 2017-05-04
Remove core_extensions from `mail` when running Ruby > 2.0

metcalf:
This PR removes all of the remaining core_extensions from mail, excluding backports from newer versions of Ruby. Each commit in this series removes a specific extension. I thought this would be easier to review as a series of commits than as a series of totally separate PRs but happy to switch.


lib/mail/core_extensions/string/access.rb
```diff
@@ -7,48 +7,6 @@
 # itself does not depend on ActiveSupport to avoid versioning conflicts
 
 class String
-  unless '1.9'.respond_to?(:force_encoding)
```
metcalf: These were unused as well. In #1094 I thought `iconv` returned a String but I confirmed that it returns an array so `first` is never called on String.

https://github.com/mikel/mail/blob/2b76d0a349895a84f02f1f86f762a4906bd133e8/lib/mail/encodings.rb#L153

lib/mail/multibyte/chars.rb
```diff
@@ -103,23 +103,23 @@ def self.wants?(string)
         # Returns a new Chars object containing the _other_ object concatenated to the string.
         #
         # Example:
-        #   ('Café'.mb_chars + ' périferôl').to_s # => "Café périferôl"
```
metcalf: Note that all the changes in this file are just rewriting comments.

jeremy: Thanks for the heads up 😊

spec/mail/utilities_spec.rb
```diff
@@ -179,13 +179,13 @@
...
       result = 'This is a string'
       expect(unbracket(test)).to eq result
     end
 
     it "should work without parens using ActiveSupport mb_chars" do
-      test = 'This is a string'.mb_chars
+      test = Mail::Multibyte.mb_chars('This is a string')
       result = 'This is a string'
       expect(unbracket(test)).to eq(result)
     end
```
jeremy: These aren't really testing ActiveSupport::Multibyte chars, huh?

lib/mail.rb
```diff
@@ -35,8 +35,6 @@ module Mail # :doc:
   if defined?(ActiveSupport)
     require 'active_support/inflector'
   else
-    require 'mail/core_extensions/string/access'
-    require 'mail/core_extensions/string/multibyte'
```
jeremy: 🤘

lib/mail/core_extensions/smtp.rb
```diff
@@ -1,25 +1,28 @@
...
-        begin
-          original_tlsconnect(s).tap { verified = true }
-        ensure
-          unless verified
-            s.close rescue nil
+# This is a backport of r30294 from ruby trunk because of a bug in net/smtp.
+# http://svn.ruby-lang.org/cgi-bin/viewvc.cgi?view=rev&amp;revision=30294
+#
+# Fixed in Ruby 1.9.3 - tlsconnect also does not exist in some early versions of ruby
+if RUBY_VERSION < '1.9.3'
```
jeremy: Note that we already do this version check in https://github.com/mikel/mail/blob/master/lib/mail.rb#L31. It's clearer to do it within this file, though! Let's drop the conditional in mail.rb.

metcalf: Changed

spec/mail/utilities_spec.rb
metcalf: Reworded these to match the current code.

jeremy:
Thanks @metcalf!!

metcalf:
Thank you! Very excited to 💥 these monkeypatches from our codebase.

Would you be willing to cut a release that includes this? If not I can cut one on our internal gem server but it's always nice to keep consistent with public releases.

jeremy:
They'll be included in the [upcoming 2.7.0 minor release](https://github.com/mikel/mail/milestone/5). I'd rather not do these removals in a 2.6.x point release.

metcalf:
Ok, thanks!

PR #1097, opened 2017-05-09
SMTP security: prevent command injection via To/From addresses

jeremy:
Validate addresses passed as SMTP command arguments to prevent
injection of other SMTP commands. Disallow line breaks and very
long addresses which may cause overflows on some old SMTP servers.

Ruby 2.4 Net::SMTP already disallows addresses that contain newlines.
Enforce this validation in Mail to cover older Ruby versions and
other SMTP implementations that don't validate input.

SMTP injection whitepaper: http://www.mbsd.jp/Whitepaper/smtpi.pdf
Ruby security report: https://hackerone.com/reports/137631
OSVDB entry: https://rubysec.com/advisories/mail-OSVDB-131677

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -1,5 +1,8 @@
 == HEAD
 
+Security:
+* #1097 – SMTP security: prevent command injection via To/From addresses. (jeremy)
+
 Features:
 * #804 - Configurable SMTP open_timeout and read_timeout. (ankane)
 * #853 - `Mail::Message#set_sort_order` overrides the default message part sort order. (rafbm)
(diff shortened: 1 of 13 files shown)


jordan-brough:
@jeremy when might we expect an official 2.6.6 release for this? Currently I only see a `2.6.6.rc1` on rubygems.

jeremy:
@jordan-brough Expect an official 2.6.6 after RC1 has had a fair number of installs to shake out regressions. Note than 2.6.x is (coincidentally) not vulnerable to this issue, thanks to #505 stripping CRLF from header values.

varyonic:
One month and >10,000 downloads for 2.6.6.rc1 so far.

jeremy:
Mail [2.5.5](https://rubygems.org/gems/mail/versions/2.5.5) and [2.6.6](https://rubygems.org/gems/mail/versions/2.6.6) are released.

PR #1102, opened 2017-05-10
Use secure protocol when running bundle install

hisas:
This PR suppresses the following warnings.
```
% bundle install
The git source `git://github.com/discourse/mini_mime.git` uses the `git` protocol, which transmits data without encryption. Disable this warning with `bundle config git.allow_insecure true`, or switch to the `https` protocol to keep your data secure.
```
Refer: bundler/bundler#4127

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,10 @@
 source 'https://rubygems.org'
 
+git_source(:github) do |repo_name|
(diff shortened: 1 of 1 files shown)


jeremy:
Thanks @hisas! Went with an explicit https source since we only bundle from GitHub for one dependency.

hisas:
I understand. Thanks for your explanation.

PR #1105, opened 2017-05-15
Use Kernel#warn instead of $stderr.puts

jeremyevans:
Kernel#warn is the standard way to issue warning messages from
ruby code.  Using it allows applications using mail to override
Kernel#warn to handle warnings in a custom manner. In Ruby 2.5+,
Kernel#warn will also call Warning.warn, unifying all ruby warning
handling.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -294,3 +294,9 @@ class Net::IMAP
     MockIMAP.new
   end
 end
+
+module Kernel
+  def warn(m)
+    $stderr.puts(m)
+  end
+end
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -294,3 +294,9 @@ def self.new(*args)
     MockIMAP.new
   end
 end
+
+module Kernel
+  def warn(m)
+    $stderr.puts(m)
+  end
+end
```
eregon: @jeremyevans Why is Kernel#warn redefined here? Shouldn't it already exist anyway?

Notably this defining is incompatible with multi-arguments `warn`.
As a result it causes https://github.com/oracle/truffleruby/issues/1900

jeremyevans: @eregon Not sure why it was added.  It should be safe to remove if it doesn't break any specs.

PR #1106, opened 2017-05-16
rfc2046

ahorek:
attached eml files can't be opened in thunderbird or outlook if they're encoded as base64

reason
https://www.ietf.org/rfc/rfc2046.txt

   It should be noted that, despite the use of the numbers "822", a
   "message/rfc822" entity isn't restricted to material in strict
   conformance to RFC822, nor are the semantics of "message/rfc822"
   objects restricted to the semantics defined in RFC822. More
   specifically, a "message/rfc822" message could well be a News article
   or a MIME message.

   No encoding other than "7bit", "8bit", or "binary" is permitted for
   the body of a "message/rfc822" entity.  The message header fields are
   always US-ASCII in any case, and data within the body can still be
   encoded, in which case the Content-Transfer-Encoding header field in
   the encapsulated message will reflect this.  Non-US-ASCII text in the
   headers of an encapsulated message can be specified using the
   mechanisms described in RFC 2047.

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -2054,6 +2054,15 @@ module Mail
       body.split!(boundary)
     end
 
+    def allowed_encodings
+      case self.mime_type
+      when 'message/rfc822'
+        [Encodings::SevenBit, Encodings::EightBit, Encodings::Binary]
+      else
+        Encodings.get_all
+      end
+    end
+
     def add_encoding_to_body
       if has_content_transfer_encoding?
         @body.encoding = content_transfer_encoding
@@ -2064,7 +2073,7 @@ module Mail
         if body && body.multipart?
             self.content_transfer_encoding = @transport_encoding
         else
-            self.content_transfer_encoding = body.get_best_encoding(@transport_encoding)
+            self.content_transfer_encoding = body.get_best_encoding(@transport_encoding, allowed_encodings)
         end
     end
 
--- a/spec/mail/message_spec.rb
+++ b/spec/mail/message_spec.rb
@@ -1401,6 +1401,22 @@ describe Mail::Message do
 
       end
 
+      # https://www.ietf.org/rfc/rfc2046.txt
+      # No encoding other than "7bit", "8bit", or "binary" is permitted for
+      # the body of a "message/rfc822" entity.
+      it "rfc2046" do
+        mail = Mail.new
+        mail.body << Mail::Part.new.tap do |part|
+          part.content_disposition = 'attachment; filename="test.eml"'
+          part.content_type  = 'message/rfc822'
+          part.body          = 'a' * 999
+        end
+        mail.encoded
+        
+        expect(mail.parts.count).to eq(1)
+        expect(mail.parts.last.content_transfer_encoding).to match(/7bit|8bit|binary/)
+      end
+
       describe "content-transfer-encoding" do
 
         it "should use 7bit for only US-ASCII chars" do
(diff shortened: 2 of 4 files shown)


lib/mail/message.rb
```diff
@@ -2063,6 +2063,8 @@ def add_encoding_to_body
     def identify_and_set_transfer_encoding
         if body && body.multipart?
             self.content_transfer_encoding = @transport_encoding
+        elsif self.mime_type == 'message/rfc822'
+            self.content_transfer_encoding = Mail::Encodings::Binary
```
jeremy: Ideally we negotiate the best of 7bit, 8bit, binary. Can we push this decisionmaking down into the existing content transfer encoding negotiations, e.g. by passing an array of allowed encodings to `body.get_best_encoding`?

spec/mail/message_spec.rb
```diff
@@ -1403,6 +1403,23 @@ def message_headers_should_match(message, other)
...
+          mail.body << Mail::Part.new.tap do |part|
+            part.header[:content_disposition] = 'attachment; filename="test.eml"'
+            part.content_type  = 'message/rfc822'
+            part.body          = 'a' * 999
+            part.body.encoding = 'UTF-8'
+          end
+
+          part_transfer_encoding = mail.to_s.scan(/Content-Transfer-Encoding: (.+)\r$/).last.first
+          puts part_transfer_encoding
+          expect(part_transfer_encoding).to match(/7bit|8bit|binary/)
```
jeremy: Can parse the encoded email and check the content transfer encoding header directly.

spec/mail/message_spec.rb
```diff
@@ -1403,6 +1403,23 @@ def message_headers_should_match(message, other)
...
+          mail = Mail.new
+          mail.body << Mail::Part.new.tap do |part|
+            part.header[:content_disposition] = 'attachment; filename="test.eml"'
+            part.content_type  = 'message/rfc822'
+            part.body          = 'a' * 999
+            part.body.encoding = 'UTF-8'
+          end
+
+          part_transfer_encoding = mail.to_s.scan(/Content-Transfer-Encoding: (.+)\r$/).last.first
+          puts part_transfer_encoding
```
jeremy: Drop debug puts

spec/mail/message_spec.rb
```diff
@@ -1403,6 +1403,23 @@ def message_headers_should_match(message, other)
...
+        # https://www.ietf.org/rfc/rfc2046.txt
+        # No encoding other than "7bit", "8bit", or "binary" is permitted for
+        # the body of a "message/rfc822" entity.
+        it "rfc2046" do
+          mail = Mail.new
+          mail.body << Mail::Part.new.tap do |part|
+            part.header[:content_disposition] = 'attachment; filename="test.eml"'
+            part.content_type  = 'message/rfc822'
+            part.body          = 'a' * 999
+            part.body.encoding = 'UTF-8'
```
jeremy: The body inherits the part's charset. Do we need to set this directly?

spec/mail/message_spec.rb
```diff
@@ -1403,6 +1403,23 @@ def message_headers_should_match(message, other)
 
       describe "content-transfer-encoding" do
 
+        # https://www.ietf.org/rfc/rfc2046.txt
+        # No encoding other than "7bit", "8bit", or "binary" is permitted for
+        # the body of a "message/rfc822" entity.
+        it "rfc2046" do
+          mail = Mail.new
+          mail.body << Mail::Part.new.tap do |part|
+            part.header[:content_disposition] = 'attachment; filename="test.eml"'
```
jeremy: We have content disposition API available here. Do we need to set the header directly?

PR #1107, opened 2017-05-17
Fix that Address#display_name and others flip-flopped between encoded/decoded formatting

### LINKED ISSUE #960: Mail::Address#display_name returns encoded or decoded value
Related to #359 and #411, the `Mail::Address#display_name` method will return either an encoded or a decoded string depending on whether `encoded` or `decoded` was called last. Reading the code, this seems very intentional (`display_name`, along with other methods, explicitly refers to `@output_type`, which is set to `:encode` or `:decode` by the `encoded` and `decoded` methods), but it's extremely counterintuitive. Why should calling `encode` on an address suddenly change the return value of `display_name`? I would expect the result of `display_name` to **always** be decoded; this method (along with `domain`, `local`, etc.) should be independent of the wire protocol.

For example, suppose you have a debug flag that conditionally logs the message being processed. Turning debugging on/off will change the output:

``` ruby
def extract_name(message, logger=nil)
  if logger
    logger.debug "Extracting address from #{message}"
  end
  message[:from].addrs.first.display_name
end

extract_name(Mail::Message.new("From: Tony =?UTF-8?B?Tm92w6Fr?= <tony@example.com>"))
# => "Tony Novák"
extract_name(Mail::Message.new("From: Tony =?UTF-8?B?Tm92w6Fr?= <tony@example.com>"), Logger.new(nil))
# => "Tony =?UTF-8?B?Tm92w6Fr?="
```

I propose doing away with `@output_type` entirely. Thoughts?

jeremy:
Due to using an `@output_type` ivar to pass state between formatting methods. Fixed by explictly passing as an argument.

Fixes #960

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -46,6 +46,9 @@ Bugs:
 * #1023 - Fix double-quoting in display names. (garethrees)
 * #1032 - Fix that comparing messages changed their raw Message-ID to their parsed message_id. (bobjflong)
 * #1074 - Fix that the first address in a list is dropped when a subsequent address has non-US-ASCII characters. (domininik)
+* #1107 - Fix Address#display_name and other formatting flip-flopping between
+encoded and decoded forms depending on whether #encoded or #decoded was called
+last. (jeremy)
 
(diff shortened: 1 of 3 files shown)


jeremy:
Cherry-picked @ d34113c

PR #1117, opened 2017-05-31
Allow overriding default Net::POP3 read_timeout value

hspazio:
**Problem:** Net::POP3 uses a default `read_timeout` of 60 seconds which can be overridden using [read_timeout=](http://ruby-doc.org/stdlib-2.0.0/libdoc/net/pop/rdoc/Net/POP3.html#method-i-read_timeout-3D) method. On an application I work we deal with mailboxes having a large volume of emails and large attachments. Sometimes, depending on the traffic we may get a `Net::ReadTimeout` error when hitting the 60 seconds limit.

This PR allows specifying the `read_timeout` configuration for Net::POP3 by passing in the `:read_timeout` configuration when specifying the retriever.

```ruby
Mail.defaults do
  retriever_method :pop3, { 
    :address             => "localhost",
    :port                => 995,
    :user_name           => "username",
    :password            => "Secret!",
    :enable_ssl          => true,
    :read_timeout        => 200 }
end
```

Regarding the spec, I used `send` to invoke a private method. I saw it being used for some other specs so I figured it could be fine. Is there a better way to test this or it's already good?

Thank you,
Fabio

### DIFF
--- a/lib/mail/network/retriever_methods/pop3.rb
+++ b/lib/mail/network/retriever_methods/pop3.rb
@@ -41,7 +41,8 @@ module Mail
                         :user_name            => nil,
                         :password             => nil,
                         :authentication       => nil,
-                        :enable_ssl           => false }.merge!(values)
+                        :enable_ssl           => false,
+                        :read_timeout         => nil }.merge!(values)
     end
     
     attr_accessor :settings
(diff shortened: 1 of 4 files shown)


hspazio:
Hello @jeremy, is there any updates on this PR? 
Thanks! ;)

PR #1119, opened 2017-06-01
Cache the expensive only_us_ascii? check

dalibor:
Is this safe to do?

In my benchmarks I'm seeing from 15-20% speed improvement when building emails with a 40KB html part and 30-40% speed improvement when building emails with large attachments.

### DIFF
--- a/lib/mail/body.rb
+++ b/lib/mail/body.rb
@@ -269,7 +269,8 @@ module Mail
     end
 
     def only_us_ascii?
-      !(raw_source =~ /[^\x01-\x7f]/)
+      return @only_us_ascii if defined?(@only_us_ascii)
+      @only_us_ascii = !(raw_source =~ /[^\x01-\x7f]/)
     end
     
     def empty?


dalibor:
@jeremy Great using `String#ascii_only?`! The only difference is that the original regex excludes `"\x00"` while `ascii_only?` includes it, does it matter? With this change the test suite time goes from 2.3 seconds to 1.6 seconds. And also, I'm seeing significant performance improvements in benchmarks locally building real emails.

```
30 attachments of ~ 10MB each:   17.801965 sec -> 5.149144 sec (~ 3x faster)
10000 html parts of ~ 40KB each: 16.148903 sec -> 9.189533 sec (~ 1x faster)
```

jeremy:
IMO that was a bug, since we don't want NUL bytes in our US-ASCII text either :)

Very nice improvements!

dalibor:
It's the other way around, the new `ascii_only?` check will allow NUL bytes in emails I'm afraid. :) If we want to prevent that, we could do `raw_source.ascii_only? && raw_source !~ /\x00/` :/, but that would be slower than `ascii_only?` and faster than the original regex.

jeremy:
Hmm yes. This seems like an implementation wart, really. NUL is legit for US-ASCII. It's not legal in emails, but that's not what we're validating here, so we can get away with the `ascii_only?`check.

dalibor:
Great, thanks for checking that, Jeremy.

Issue #1143, opened 2017-09-06
Message#attachments should count up singlepart attachment file.


kirikak2:
In singlepart mail with attachment file, Mail library cannot count attachment file.

This is a sample case.

```
mail = Mail.new(File.read("spec/fixtures/emails/attachment_emails/attachment_only_email.eml"))
mail.attachments.count
=> 0
```

I expect mail.attachments.count retuns `1`.

PR #1144, opened 2017-09-06
Fix that Message#attachemnts cannot count singlepart attachment

kirikak2:
see #1143


lib/mail/message.rb
```diff
@@ -2167,5 +2171,19 @@ def do_delivery
...
       Encodings.transcode_charset decode_body, charset, 'UTF-8'
     end
+
+    def generate_attachment_part_from_body
+      part = Mail::Part.new(nil)
+      header_part = []
+      header_part << header[:content_type].encoded if header[:content_type]
+      header_part << header[:content_disposition].encoded if header[:content_disposition]
+      header_part << header[:content_transfer_encoding].encoded if header[:content_transfer_encoding]
+      header_part << header[:content_discription].encoded if header[:content_discription]
```
evaldasg: `header[:content_description]` typo here

HashNotAdam:
Unfortunately, this issue is causing me problems when parsing DMARC reports coming from Google. I've tested this branch and, apart from correcting the above-mentioned typo, found that it worked without modification.

@kirikak2 any chance you can get this over the line? Alternatively, I could make a new PR based on this one but compatible with the current version of mail

PR #1145, opened 2017-09-07
Fix that set not-accurate multipart header

kirikak2:
According to RFC2046, multipart/* content-type do not need to have charset parameter.
Currentry, When I create multipart mail, mail library append `charset="UTF-8"` in content-type header automatically.

This patch fixes 2 issues.

1. When create multipart content-type header, mail library appends charset parameter.
2. When call `Message#convert_to_multipart`, mail library ignores original header. I think `Content-Type`, `Content-Disposition`, `Content-Transfer-Encoding` and`Content-Description` should move to part header.


jeremy:
Agreed on both points.

Issue #1038, opened 2016-10-05
Message#from_yaml results in invalid parts when body is not present


angelcarballo:
Found this behaviour while testing serialization and deserialization of messages via YAML.

The following spec fails (at least on `master` and `2.6.4`)

``` ruby
context 'when body is not present' do
  it 'should encode after deserializing' do
    message = Mail::Message.new(
      to: 'someone@somewhere.com',
      cc: 'someoneelse@somewhere.com',
      bcc: 'someonesecret@somewhere.com',
      subject: 'subject'
    )
    serialized = message.to_yaml
    deserialized = Mail::Message.from_yaml(serialized)
    expect{ deserialized.encoded }.not_to raise_error
  end
end
```

The `#from_yaml` method sets `@body` directly, which after YAML serialization will contain an Array on `parts` instead of a `PartList` instance.

Is this a bug, or it's the expected behaviour (making the body mandatory) ?

jeremy:
```ruby
>> Mail::Message.from_yaml(Mail.new.to_yaml).encoded
ArgumentError: wrong number of arguments (given 1, expected 0)
	from /Users/jeremy/source/mail/lib/mail/message.rb:1801:in `sort!'
```
ಠ_ಠ

jeremy:
Caused by #782

dragosmiron:
still failing on 2.7.0 @jeremy

jeremy:
@dragosmiron Expected; thanks for the report. Pull requests welcome 😊

dragosmiron:
But the fix is above in the PR, and in master, just not in a release yet. Any idea when the next release might be? We had to fork the gem to benefit from your 'master' code

jeremy:
@dragosmiron Got a link to the fix you're referencing?

dragosmiron:
https://github.com/mikel/mail/pull/1148

jeremy:
> But the fix is above in the PR, and in master, just not in a release yet.

#1148 was not merged to master. The PR is a workaround for a symptom of the problem, not a fix for the underlying issue, so it was closed as a duplicate of this issue.

dragosmiron:
cool, https://github.com/mikel/mail/pull/1148 is good enough until this is somehow fixed, so we had to fork the gem.

dragosmiron:
hello, two years later, has this issue or the one from https://github.com/mikel/mail/pull/1148 been fixed?

dragosmiron:
[...]

PR #1148, opened 2017-09-07
Fix "wrong number of arguments (given 1, expected 0)" in "sort!"

rivsc:
…. And Array#sort! doesn't take arguments.

wrong number of arguments (given 1, expected 0)

### Backtrace ### :
	/Users/xxx/mail/lib/mail/body.rb:133:in `sort!'
	/Users/xxx/mail/lib/mail/body.rb:133:in `sort_parts!'
	/Users/xxx/mail/lib/mail/body.rb:131:in `block in sort_parts!'
	/Users/xxx/.rvm/rubies/ruby-2.4.0/lib/ruby/2.4.0/delegate.rb:341:in `each'
	/Users/xxx/.rvm/rubies/ruby-2.4.0/lib/ruby/2.4.0/delegate.rb:341:in `block in delegating_block'
	/Users/xxx/mail/lib/mail/body.rb:129:in `sort_parts!'
	/Users/xxx/mail/lib/mail/body.rb:153:in `encoded'

### DIFF
--- a/lib/mail/body.rb
+++ b/lib/mail/body.rb
@@ -130,7 +130,7 @@ module Mail
         p.body.set_sort_order(@part_sort_order)
         p.body.sort_parts!
       end
-      @parts.sort!(@part_sort_order)
+      @parts.sort!(@part_sort_order) if @parts.kind_of?(Mail::PartsList)
     end
     
     # Returns the raw source that the body was initialized with, without
--- a/spec/mail/body_spec.rb
+++ b/spec/mail/body_spec.rb
@@ -457,4 +457,12 @@ describe Mail::Body do
       expect(body.encoded).to eq 'The Body'
     end
   end
+  
+  describe "Partslist empty" do
+    it "should not break on empty PartsList on body" do
+      body = Mail::Body.new('The Body')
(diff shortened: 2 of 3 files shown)


jeremy:
Nice catch. Test case?

rivsc:
I have added test case. I hope that is useful.

bretweinraub:
I'm getting this same error from 

Rails 5.1.4/ActionMailer 5.1.4, but only when delivering via a delayed job.

First, this was stack tracing:

https://github.com/mikel/mail/blob/master/lib/mail/message.rb#L1658

    def attachments
      parts.attachments
    end

And I replaced this with something a little more circumspect:

    def attachments
      parts && parts.respond_to?(:attachments) ? parts.attachments : (parts || [])          
    end

And now I get the error as described in this ticket.

Oddly, outside of delayed job, everything works swimmingly.

jeremy:
Duplicate of #1038 👍

Caused by serializing to YAML, deserializing from YAML, then encoding the message. `@parts` is deserialized as an Array instead of a PartsList. Caused by #782.

PR #1159, opened 2017-10-24
Parse emails with \n newlines if they don't contain binary content

### LINKED ISSUE #1129: LF emails can't be parsed anymore
removed in #1113 / master branch

I have many eml files in my specs with just LF separator, that passes on 2.6.6 stable
```
HEADER
LFLF
BODY
```

when normalization was removed, header separator doesn't work with it anymore, the parser thinks that the whole email is a header which is wrong.
```
HEADER_SEPARATOR = /#{CRLF}#{CRLF}/
```

according to the RFC, header separator should be always 2x CRLF
https://www.ietf.org/rfc/rfc2822.txt

so my question is, should eml files with just LF separators be parsable or not?
It doesn't comply with RFC, but standard mail clients like Outlook or Thunderbird can open and work with these files just fine. It also used to work with mail 2.6.6.

jeremy:
Binary-safe newline conversion ensures that attachments don't get corrupted. Requires that text is ASCII-only.

Fixes #1129
References #1113

### DIFF
--- a/lib/mail/encodings/7bit.rb
+++ b/lib/mail/encodings/7bit.rb
@@ -4,29 +4,19 @@ require 'mail/encodings/8bit'
 
 module Mail
   module Encodings
+    # 7bit and 8bit are equivalent. 7bit encoding is for text only.
     class SevenBit < EightBit
       NAME = '7bit'
-    
       PRIORITY = 1
+      Encodings.register(NAME, self)
 
-      # 7bit and 8bit operate the same
-      
-      # Decode the string
       def self.decode(str)
-        super
-      end
-    
-      # Encode the string
-      def self.encode(str)
-        super
-      end
-     
-      # Idenity encodings have a fixed cost, 1 byte out per 1 byte in
-      def self.cost(str)
-        super 
+        ::Mail::Utilities.binary_unsafe_to_lf str
       end
 
-      Encodings.register(NAME, self) 
+      def self.encode(str)
+        ::Mail::Utilities.binary_unsafe_to_crlf str
+      end
     end
   end
 end
--- a/lib/mail/encodings/8bit.rb
+++ b/lib/mail/encodings/8bit.rb
@@ -6,32 +6,13 @@ module Mail
   module Encodings
     class EightBit < Binary
       NAME = '8bit'
-
       PRIORITY = 4
-
-      # 8bit is an identiy encoding, meaning nothing to do
-      
-      # Decode the string
-      def self.decode(str)
-        ::Mail::Utilities.to_lf str
-      end
-    
-      # Encode the string
-      def self.encode(str)
-        ::Mail::Utilities.to_crlf str
-      end
-     
-      # Idenity encodings have a fixed cost, 1 byte out per 1 byte in
-      def self.cost(str)
-        1.0
-      end
+      Encodings.register(NAME, self)
 
       # Per RFC 2821 4.5.3.1, SMTP lines may not be longer than 1000 octets including the <CRLF>.
       def self.compatible_input?(str)
-        !str.lines.find { |line| line.length > 998 }
+        !str.lines.find { |line| line.bytesize > 998 }
       end
-
-      Encodings.register(NAME, self) 
     end
   end
 end
--- a/lib/mail/encodings/base64.rb
+++ b/lib/mail/encodings/base64.rb
@@ -4,36 +4,35 @@ require 'mail/encodings/7bit'
 
 module Mail
   module Encodings
+    # Base64 encoding handles binary content at the cost of 4 output bytes
+    # per input byte.
     class Base64 < SevenBit
       NAME = 'base64'
-
       PRIORITY = 3
+      Encodings.register(NAME, self)
 
       def self.can_encode?(enc)
         true
       end
 
-      # Decode the string from Base64
       def self.decode(str)
-        RubyVer.decode_base64( str )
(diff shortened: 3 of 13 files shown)


lib/mail/encodings/7bit.rb
```diff
@@ -4,29 +4,19 @@
...
-      # Encode the string
+
       def self.encode(str)
-        super
-      end
-     
-      # Idenity encodings have a fixed cost, 1 byte out per 1 byte in
-      def self.cost(str)
-        super 
+        ::Mail::Utilities.binary_unsafe_to_crlf str
```
jeremy: 7bit encoding → never binary

lib/mail/encodings/base64.rb
```diff
@@ -4,36 +4,35 @@
...
-      # Decode the string from Base64
       def self.decode(str)
-        RubyVer.decode_base64( str )
+        RubyVer.decode_base64(str)
       end
 
-      # Encode the string to Base64
       def self.encode(str)
-        ::Mail::Utilities.to_crlf(RubyVer.encode_base64( str ))
+        ::Mail::Utilities.binary_unsafe_to_crlf(RubyVer.encode_base64(str))
```
jeremy: Base64 is 7bit ASCII, ok to convert newlines

lib/mail/encodings/8bit.rb
```diff
@@ -6,32 +6,13 @@ module Mail
...
-      
-      # Decode the string
-      def self.decode(str)
-        ::Mail::Utilities.to_lf str
-      end
-    
-      # Encode the string
-      def self.encode(str)
-        ::Mail::Utilities.to_crlf str
-      end
```
jeremy: No longer performing newline conversion in 8bit/binary encodings

lib/mail/network/delivery_methods/sendmail.rb
```diff
@@ -61,7 +61,7 @@ def deliver!(mail)
 
     def self.call(path, arguments, destinations, encoded_message)
       popen "#{path} #{arguments} #{destinations}" do |io|
-        io.puts ::Mail::Utilities.to_lf(encoded_message)
+        io.puts ::Mail::Utilities.binary_unsafe_to_lf(encoded_message)
```
jeremy: Encoded message is 7bit ASCII, safe to convert newlines for `sendmail` (which expects LF newlines)

lib/mail/utilities.rb
```diff
@@ -248,33 +248,62 @@ def map_with_index( enum, &block )
...
+    # encoding 8bit and base64 Content-Transfer-Encoding and for convenience
+    # when parsing emails with \n line endings instead of the required \r\n.
+    def self.to_crlf(string)
+      string = string.to_s
+      if safe_for_line_ending_conversion? string
+        binary_unsafe_to_crlf string
+      else
+        string
+      end
     end
```
jeremy: `to_lf` and `to_crlf` are now "best attempt" converters. If the string is unsafe for conversion, they aren't modified.

spec/mail/body_spec.rb
```diff
@@ -443,7 +443,7 @@ def assert_split_into(body, pre, epi, parts)
 
   describe "non US-ASCII charset" do
     it "should encoded" do
-      body = Mail::Body.new("あいうえお\n")
+      body = Mail::Body.new("あいうえお\r\n")
```
jeremy: Since non-ASCII chars are seen as possibly-binary. We could improve the heuristic to check for other valid text encodings to allow this.

dalibor:
@jeremy Thanks for the recent bug fixes and the 2.7.0.rc2 release! I ran my test suite with it and the only thing I'm seeing are few examples where this LF to CRLF conversion is not applied when parsing emails (see below). Using the `Mail::Utilities.binary_unsafe_to_crlf` helper I'm able to fix my email source fixtures, so that's good.

1) Header / body CRLF split by `Mail::Message::HEADER_SEPARATOR`

```ruby
$ irb
>> gem 'mail', '2.6.6'; require 'mail'
=> true
>> Mail.new("Subject: Subject\n\nContent").body.decoded
=> "Content"
```

vs

```ruby
$ irb
>> gem 'mail', '2.7.0.rc2'; require 'mail'
=> true
>> Mail.new("Subject: Subject\n\nContent").body.decoded
=> ""
```

2) Using mail parts:

```ruby
$ irb
>> gem 'mail', '2.6.6'; require 'mail'
=> true
>> m1 = Mail.new; m1.text_part = Mail::Part.new('text'); m2 = Mail.new(m1.to_s.gsub("\r\n", "\n")); m2.text_part.decoded
=> "text"
```

vs

```ruby
$ irb
>> gem 'mail', '2.7.0.rc2'; require 'mail'
=> true
>> m1 = Mail.new; m1.text_part = Mail::Part.new('text'); m2 = Mail.new(m1.to_s.gsub("\r\n", "\n")); m2.text_part.decoded
NoMethodError: undefined method `decoded' for nil:NilClass
```

It fails because no text part was detected due to no proper CRLF in that source.

Not sure if this PR is intended to cover the above two cases, just wanted to bring it up here.

Thanks!

jeremy:
Thanks @dalibor! I'll look at supporting these cases too 👍

jeremy:
@dalibor fixed @ #1160

dalibor:
@jeremy Thanks, latest patch works perfectly!

I'll do more tests in prod with 2.7.0.rc3 in the following days.

PR #1168, opened 2017-11-03
restore LF->CRLF conversion for properly encoded non-binary messages

rubys:
Useful for templated emails and test data

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -10,6 +10,8 @@ Compatibility:
 Performance:
 * Speed up message encoding, especially with large attachments. (dalibor)
 
+Features:
+* Restore LF=>CRLF conversions for properly encoded non-binary emails (rubys)
 
 == Version 2.7.0 (2017-10-31)
 
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -1962,7 +1962,6 @@ module Mail
     end
 
     def raw_source=(value)
-      value = value.dup.force_encoding(Encoding::BINARY) if RUBY_VERSION >= "1.9.1"
       @raw_source = ::Mail::Utilities.to_crlf(value)
     end
 
--- a/lib/mail/utilities.rb
+++ b/lib/mail/utilities.rb
@@ -258,7 +258,12 @@ module Mail
     end
 
     def self.safe_for_line_ending_conversion?(string) #:nodoc:
-      string.ascii_only?
+      if RUBY_VERSION >= '1.9'
+        string.ascii_only? or 
+          (string.encoding != Encoding::BINARY and string.valid_encoding?)
+      else
+        string.ascii_only?
+      end
     end
 
     # Convert line endings to \n unless the string is binary. Used for
--- a/spec/mail/header_spec.rb
+++ b/spec/mail/header_spec.rb
@@ -460,6 +460,20 @@ HERE
       expect(header['Received'].value).to eq "from [127.0.220.158] (helo=fg-out-1718.google.com)\tby smtp.totallyrandom.com with esmtp (Exim 4.68)\t(envelope-from <stuff+caf_=support=aaa.somewhere.com@gmail.com>)\tid 1K4JeQ-0005Nd-Ij\tfor support@aaa.somewhere.com; Thu, 05 Jun 2008 10:53:29 -0700"
     end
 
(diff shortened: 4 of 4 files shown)


jeremy:
Thanks @rubys 😊

jeremy:
Merged @ dedb669573b672e8dedeaab42e9515af5c01a45d and backported to 2-7-stable @ d17be66765f7f63caa5d5cedb67a597f853a4402

leez:
pleeeease release 2.7.1! This bug is driving us crazy. Thanks :)

uncleramsay:
Is this change going to be released any time soon?

jeremy:
Yep @uncleramsay.

SamSaffron:
just putting out there that we can not upgrade Discourse to this gem yet due to this bug, hoping we can get a release out soon, is there anything I can do to help?

jeremy:
Note that this is not a bug—you can work around invalid line endings (e.g. `\n` in fixtures, etc) by explicitly converting them to valid ones with `Mail::Utilities.binary_unsafe_to_crlf(…)`.

2.7.1 will obviate this for some messages, but not all of them. No need to wait on it otherwise!

SamSaffron:
Oh, will see what we can do to fix our test suite not to depend on this old behavior I guess.

rubys:
@jeremy: two things:

1) I believe that `binary_unsafe_to_crlf` may corrupt attachments (the original problem that was corrected that caused this problem).

2) I subsequently found another fix that will safely handle considerably more improperly formatted emails:

```
diff --git a/lib/mail/message.rb b/lib/mail/message.rb
index 5994251..8d31f54 100644
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -1956,7 +1956,7 @@ module Mail
 
   private
 
-    HEADER_SEPARATOR = /#{Constants::CRLF}#{Constants::CRLF}/
+    HEADER_SEPARATOR = /\r?\n\r?\n/
 
     #  2.1. General Description
     #   A message consists of header fields (collectively called "the header
```

This change does not appear to cause any more existing tests to fail (at the moment, I'm seeing three failures with or without this change).

An example of an improperly formatted email that can be successfully parsed:

```
Mail.new("Subject: hi there\nfoo: bar\n\nIñtërnâtiônàlizætiøn".force_encoding('binary')).to_s
```

Let me know if you are interested in a pull request along these lines.

SamSaffron:
@jeremy well, we tried upgrading and had to revert cause somehow it cased our ruby heaps to grow, will investigate this separately and see if I can make a report of what causes this. 

image

jeremy:
@SamSaffron looking forward to that, esp. if you can bisect 😋

SamSaffron:
I need some magic plan for bisecting this stuff in PRD, I actually have a bunch of other experiments I want to run that would make use of a nice "bisect in PRD" kind of grafana dashboard where we run multiple apps in different versions concurrently. 

In theory we could spin up like 10 AWS instances all with different commits and then just watch for a day.

jeremy:
@rubys Pursued that approach in #1208. Split headers/body on any permissible line ending pair, then to_crlf on each of them afterward.

jeremy:
Mail 2.7.1.rc1 released. Please take it for a spin.

TomFreudenberg:
Hi @jeremy 

I am running into this issue also a number of times with 2.7.0. Do you already know a date for publishing 2.7.1?

Thanks for your work
Tom

jeremy:
@TomFreudenberg Have you tried 2.7.1.rc1? Weighing whether to loosen the current rules (ascii_only check) further.

TomFreudenberg:
Hi @jeremy No, I haven't tried the rc1 but applied once this commit on my 2.7.0. That works fine for me.

TomFreudenberg:
Hi @jeremy 

just to let you know, we have _corrected_ that in our MidiSmtpServer SMTPD library. So for now we are transforming incoming message data into RFC (2)822 conform message data. That means, while receinving incoming messages, we make sure that each line ends up by CRLF. For then Mail does work also in current releases. Maybe checkout our issue 16 about "Handling of CRLF and LF - RFC 2822 STRICT or BEST PRACTISE".

Thanks for your work and help
Tom

furuhama:
thanks!

TomFreudenberg:
@jeremy thanks for update and information!

PR #1172, opened 2017-11-08
Update mini_mime dependency to 1.0.0

radar:
(Assuming https://github.com/discourse/mini_mime/pull/9 goes through)

The `mini_mime` gem version dependency is open ended enough that a possible future `mini_mime` release with breaking API changes _could_ potentially break current `mail` gem versions. 

To prevent this from happening, we should lock down the `mini_mime` dependency, according to semver best practices.


jeremy:
Premature semver lockdown sounds great on paper yet causes dep hell in practice, particularly with library deps like these. Happy to put down the screws (limit supported mini_mime versions) in a stable mail release if mini_mime does oddly go bananas!

PR #1180, opened 2017-12-08
Use "Re: " instead of "RE: "

mashedcode:
As specified in [RFC 5322 3.6.5.](https://tools.ietf.org/html/rfc5322#section-3.6.5)


jeremy:
Dig it. Thank you @mashedcode.

PR #1192, opened 2017-12-28
remove jruby hack

ahorek:
it looks like it's already fixed in the current version of jruby


jeremy:
Yes indeed. Thanks @ahorek!

PR #1194, opened 2018-01-10
Reduce warnings

utilum:
Ruby 2.5 warns about these.

### DIFF
--- a/lib/mail/parsers/address_lists_parser.rb
+++ b/lib/mail/parsers/address_lists_parser.rb
@@ -31992,7 +31992,6 @@ end
 
       
 begin
-	testEof = false
 	_slen, _trans, _keys, _inds, _acts, _nacts = nil
 	_goto_level = 0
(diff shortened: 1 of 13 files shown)


lib/mail/parsers/address_lists_parser.rb
```diff
@@ -31992,7 +31992,6 @@ def self.parse(data)
 
       
 begin
-	testEof = false
```
jeremy: `lib/mail/parsers/*_parser.rb` are generated from `.rl` files by Ragel—this is boilerplate generated Ruby code that we can't change. Hence the `$VERBOSE = nil` wrapping the parser definition, suppressing warnings 😑

utilum: Oh... we do see the warnings running gems that require mail, but it seems that correcting this should be done elsewhere. Sorry for the noise.

PR #1198, opened 2018-01-29
Fix parsing boundary containing "=" within invalid Content-Type

rafbm:
The following header was see in the wild:

```
Content-Type: multipart/mixed;
 boundary="=_NextPart_2rfkindysadvnqw3nerasdf";windows-852
```

Due to the value-less `windows-852` param at the end, this header raises in the parser and gets sanitized before being re-parsed. The sanitize method, among other things, splits on "=" in order to reconstruct key-value params. Because the boundary contains "=" itself, everything after "=" was ignored.

Resulting params before commit:

```ruby
{ "boundary" => "\\\"", "windows-852" => "" }
```

Resulting params after commit:

```ruby
{ "boundary" => "=_NextPart_2rfkindysadvnqw3nerasdf", "windows-852" => "" }
```

Params are key-value pairs, so it just makes sense to split at most 2 parts.

### DIFF
--- a/lib/mail/fields/content_type_field.rb
+++ b/lib/mail/fields/content_type_field.rb
@@ -150,7 +150,7 @@ module Mail
         # and: audio/x-midi;\r\n\sname=Part .exe
         params = $2.to_s.split(/\s+/)
         params = params.map { |i| i.to_s.chomp.strip }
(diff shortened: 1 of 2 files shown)


jeremy:
Nice. Thanks @rafbm!

PR #1199, opened 2018-01-30
Backport 2.7-stable fix to 2.6-stable for quoted-printable and CR/LF problems

muhammadn:
Backported 2.7-stable for correct handling of quoted-printable.

Currently 2.6-stable has a bug with attachments being corrupted by CR/LF normalization.

As we cannot use 2.7.0 for our needs because in our tests, 2.7.0's rewrite of some core code makes our tests fail, so we had no choice but to backport the fixes for QP (not core 2.7 code) to 2.6-stable.

We had backported this commit from 2.7-stable: https://github.com/mikel/mail/commit/21222e1b48f08f6f848d26643d6b6fb1d873d18c


ahorek:
IMO this shouln't be backported, because it's a breaking change. Can you investigate why are your tests failing on 2.7.0?

jeremy:
Yep, stable branch = no backports. Thanks for the PR, though!

PR #1210, opened 2018-03-02
Normalize Quoted-Printable line endings for text content

jeremy:
If a string has binary content, leave \r\n characters as-is.

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -9,6 +9,7 @@ Breaking changes:
 
 Compatibility:
 * Handle a wide variety of non-RFC Message-ID formats. (peterkovacs)
+* Normalize Quoted-Printable line endings for text content. (jeremy)
 * Gracefully parse invalid dates in Date and Received headers. (okkez)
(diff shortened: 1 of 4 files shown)


chuckremes2:
What is the status of this PR? Will it be merged for a `2.7.2` release at some point?

Uysim:
when we can merge this?

jemminger:
One year later...is there a status on this?

dcormier:
For reference, from [RFC 2045, section 6.7](https://tools.ietf.org/html/rfc2045#section-6.7):

> An "=" followed by a character that is neither a hexadecimal digit (including "abcdef") nor the CR character of a CRLF pair is illegal.

So, having an `=` followed by a `\n` (linefeed, also known as LF) is _not_ legal in quoted-printable content. It looks like this gem currently encodes content such that it may result in an illegal `=\n` character sequence. This PR looks like it does fix that.

jirihradil:
Hi guys, any update for 2.7.2, please?

PR #1218, opened 2018-03-16
Prevent “invalid byte sequence in UTF-8” when initializing from string

rafbm:
When upgrading to current master, we had regressions in our test suite for all emails containing non-valid UTF-8 bytes, which are often seen in the wild. The exception was raised by `String#match` in `#set_envelope_header`, which is called right after `#raw_source=` in `#init_with_string`.

Bringing back `#force_encoding` in `#raw_source=` (removed in dedb6695) solved all of these without breaking any other spec, so I figured it was a sensible fix. Let me know if something else would be more appropriate.

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -1981,6 +1981,7 @@ module Mail
     end
 
     def raw_source=(value)
+      value = value.dup.force_encoding(Encoding::BINARY) if RUBY_VERSION >= "1.9.1"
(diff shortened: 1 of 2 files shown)


rafbm:
Closing this as it breaks some real-life emails. I’ll try to take the time to contribute specs / fixtures that would have been broken by this change.

jeremy:
Thanks @rafbm!

Issue #1226, opened 2018-04-16
Forcing Content-Transfer-Encoding for a mail part is difficult


skalee:
Some MIME types explicitly allow only a subset of possible transport encodings. For example, RFC 1847 "Security Multiparts" defines a "Multipart/Signed" MIME type, which [requires that its first component has transport encoding quoted-printable or base64](https://tools.ietf.org/html/rfc1847#section-2.1 ). Hence, under some circumstances, default encodings guessed by this gem are incorrect, and users have well-advocated need to override the transport encoding.

This gem's public API seems to provide three ways of setting content transfer encoding. Neither is documented, nor at least mentioned in README:

1. `Message#content_transfer_encoding=` — seems to have no permanent effect, and is reset on `Message#encoded` call
2. `Message#transfer_encoding=` — unusable for multipart messages, because inner part encoding is reset by `Message#encoded` call on the outer part
3. `Body#encoding=` — worked best in my case

Perhaps that many of encoding setters are necessary, however more documentation is really needed. At minimum, there should be some mention in README which hints the best way of forcing transport encoding.

For reference, my Pry session which shows the behaviour of these three methods is recorded in this Gist: https://gist.github.com/skalee/f724b3c69657aa0b836a9208b2505af1

ahorek:
Hello,
I agree with @jeremy. IMO transfer encoding setters shouln't be public. Such API could be done, but right now these methods are used only for internal purposes and I don't see any real use case except testing.

_Message#content_transfer_encoding=_
this example should enforce the encoding (even if it violates standards) or it's just a prefered encoding?
 
There could be a case when a generated mail violates some RFC because of the wrong transport encoding. This should be adressed and fixed, especially if it causes real problems (the email could be rejected by MX or unreadable).
Could you share a minimal example, that demonstrates such behaviour? It shouldn't be hard to fix.

skalee:
Hi @ahorek!

Thank you for your response.

Some software is known to modify/convert messages in a subtle way (e.g by converting the transfer encoding to 7bit, quoted-printable, or base64, or removing trailing spaces). It doesn't matter for many users because the meaning of the message is preserved. The message will be delivered, and will be readable. That said, when you're going to sign your message with Open PGP (and probably some other tools as well), you got another problem: such alterations typically invalidate the signature.

[...]

PR #1227, opened 2018-04-18
WIP: fix RFC 1847 / Multipart/Signed

ahorek:
fixes https://github.com/mikel/mail/issues/1226

### DIFF
--- a/spec/mail/message_spec.rb
+++ b/spec/mail/message_spec.rb
@@ -1432,9 +1432,24 @@ describe Mail::Message do
           part.body          = 'a' * 999
         end
         mail.encoded
-        
+
         expect(mail.parts.count).to eq(1)
         expect(mail.parts.last.content_transfer_encoding).to match(/7bit|8bit|binary/)
+      end
+
+      # https://www.ietf.org/rfc/rfc1847.txt
+      # No encoding other than "quoted-printable", or "base64" is permitted for
+      # the body of a "multipart/signed" entity.
+      it "rfc1847" do
+        mail = Mail.new
+        mail.body << Mail::Part.new.tap do |part|
+          part.content_disposition = 'attachment; filename="test.asc"'
+          part.content_type  = 'application/pgp-signature'
+        end
+        mail.encoded
+
+        expect(mail.multipart?).to eq(true)
+        expect(mail.parts.last.content_transfer_encoding).to match(/quoted-printable|base64/)
       end
 
       describe "convert_to_multipart" do
--- a/spec/mail/message_spec.rb
+++ b/spec/mail/message_spec.rb
@@ -1448,7 +1448,7 @@ describe Mail::Message do
         end
         mail.encoded
 
-        expect(mail.multipart?).to eq(true)
+        expect(mail.multipart?).to be_truthy
         expect(mail.parts.last.content_transfer_encoding).to match(/quoted-printable|base64/)
       end
 
--- a/lib/mail/encodings/transfer_encoding.rb
+++ b/lib/mail/encodings/transfer_encoding.rb
@@ -36,11 +36,17 @@ module Mail
         message_encoding = Encodings.get_encoding(message_encoding) || Encodings.get_encoding('8bit')
         source_encoding  = Encodings.get_encoding(source_encoding)
 
-        if message_encoding && source_encoding && message_encoding.can_transport?(source_encoding) && source_encoding.compatible_input?(str)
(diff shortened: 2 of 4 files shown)


spec/mail/message_spec.rb
```diff
@@ -1432,11 +1432,26 @@ def message_headers_should_match(message, other)
...
+      # the body of a "multipart/signed" entity.
+      it "rfc1847" do
+        mail = Mail.new
+        mail.body << Mail::Part.new.tap do |part|
+          part.content_disposition = 'attachment; filename="test.asc"'
+          part.content_type  = 'application/pgp-signature'
+        end
+        mail.encoded
+
+        expect(mail.multipart?).to eq(true)
```
jeremy: `be_truthy`

rafbm: @jeremy Out of curiosity, do you prefer `be_truthy` just for consistency across the project, a personal taste, or to leave room for `multipart?` returning a truthy object that isn’t `true`?

jeremy: All of the above, yes! In essence, follow Ruby's lead that implicit coercion to boolean is a desirable default. Use explicit boolean values when explicitly necessary, which is rare.

skalee:
Thanks a lot, but no, it doesn't fix the issue.

The `application/pgp-signature` part may have any transfer encoding. The signature itself should be resilent to encoding changes — as are images, videos, or audios. Furthermore, signatures are typically in textual format. For instance, Open PGP encodes signature in it's own way called ASCII Armor (roughly base64, just different chars are used), and such signature can be transferred with any encoding, including 7bit, without any loss.

----

A multipart/signed message is structured as follows:

    container part or message (multipart/signed)
      |-> part which is signed (any mime type, incl. multipart)
      |-> signature (application/pgp-signature in case of Open PGP,
                     application/pkcs7-signature in case of S/MIME,
                     possibly others as well)

Content-Transfer-Encoding restrictions apply to the part which is signed, not to the signature.  It is required by standard that it's a first subpart of multipart/signed.  However, the MIME type of that part can be anything.

The reason why transfer encoding is important is that the signature is calculated over encoded message part.  The message composer first encodes the part which is about to be signed, then calculates a signature over that encoded part, and finally builds a multipart/signed.  One of the benefits of this behaviour is that it guarantees the integrity of MIME headers as well, esp. Content-Type and Content-ID.

Following example message comes from https://tools.ietf.org/html/rfc3156#section-5, and describes what happens:

      From: Michael Elkins <elkins@aero.org>
      To: Michael Elkins <elkins@aero.org>
      Mime-Version: 1.0

      Content-Type: multipart/signed; boundary=bar; micalg=pgp-md5;
        protocol="application/pgp-signature"

      --bar
    & Content-Type: text/plain; charset=iso-8859-1
    & Content-Transfer-Encoding: quoted-printable
    &
    & =A1Hola!
    &
    & Did you know that talking to yourself is a sign of senility?
    &
    & It's generally a good idea to encode lines that begin with
    & From=20because some mail transport agents will insert a greater-
    & than (>) sign, thus invalidating the signature.
    &
    & Also, in some cases it might be desirable to encode any   =20
    & trailing whitespace that occurs on lines in order to ensure  =20
    & that the message signature is not invalidated when passing =20
    & a gateway that modifies such whitespace (like BITNET). =20
    &
    & me

      --bar

      Content-Type: application/pgp-signature

      -----BEGIN PGP MESSAGE-----
      Version: 2.6.2

      iQCVAwUBMJrRF2N9oWBghPDJAQE9UQQAtl7LuRVndBjrk4EqYBIb3h5QXIX/LC//
      jJV5bNvkZIGPIcEmI5iFd9boEgvpirHtIREEqLQRkYNoBActFBZmh9GC3C041WGq
      uMbrbxc+nIs1TIKlA08rVi9ig/2Yh7LFrK5Ein57U/W72vgSxLhe/zhdfolT9Brn
      HOxEa44b+EI=
      =ndaj
      -----END PGP MESSAGE-----

      --bar--

With following note:

> The "&"s in the previous example indicate the portion of the data over which the signature was calculated.

If some gateway would transform the signed part to base64 or quoted-printable, it would produce a different sequence of bytes, making signature invalid.  Parts encoded as quoted-printable or base64 are never altered during transport, hence are safe to use.

----

BTW I need to check what transfer encoding should be used when signing a multipart part.

ahorek:
thanks for explanation! let me will fix it

PR #1228, opened 2018-04-18
optimize unwraps

ahorek:
benchmark
```
without braces
patch:   3199358.0 i/s
master:  1166684.9 i/s - 2.74x  slower

with braces
patch:   2222942.3 i/s
master:   485430.5 i/s - 4.58x  slower
```


ahorek:
ping @jeremy ?

I think it's much simplier and more readable than the original regex version. Let me know if you have any concerns.

PR #1229, opened 2018-04-18
improve performance of Message#find_attachment

ahorek:
this is a rebased version of https://github.com/mikel/mail/pull/825

benchmark
```ruby
require 'mail'
require 'benchmark/ips'
part = Mail::Part.new.tap do |part|
  part.content_disposition = 'attachment; filename="test.asc"'
  part.content_type  = 'application/pgp-signature'
end

Benchmark.ips do |x|
  x.report('test find_attachment') { part.send(:find_attachment) }
end
```

result
```
patch:     25437.0 i/s
master:    13268.2 i/s - 1.92x  slower
```

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -2114,21 +2114,30 @@ module Mail
 
     # Returns the filename of the attachment (if it exists) or returns nil
     def find_attachment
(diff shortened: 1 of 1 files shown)


ahorek:
rebased

PR #1232, opened 2018-04-21
use unpack1 if available

ahorek:
https://github.com/JuanitoFatas/fast-ruby#stringunpack1-vs-stringunpack0-code


ahorek:
moved to https://github.com/mikel/mail/pull/1513

PR #1233, opened 2018-04-21
respect allowed encodings

ahorek:
we should respect allowed encodings, even if the original encoding is wrong

what do you think about it @mikel? #1226


ahorek:
because this situation can't happen with the current code base, I'm closing it for now

PR #1242, opened 2018-06-11
Use spelling for MIME-Version as defined in RFC 2045

paroga:
This will result in better rating by some spam checkers.

e.g.: [MV_CASE](https://github.com/vstakhov/rspamd/blob/0e654a69f908ae3abe19663dc192f1dbc45d8ed0/rules/headers_checks.lua#L561) rule in Rspamd

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,14 +418,14 @@ Content-Type: multipart/alternative;
   boundary=--==_mimepart_4a914f0c911be_6f0f1ab8026659
 Message-ID: <4a914f12ac7e_6f0f1ab80267d1@baci.local.mail>
 Date: Mon, 24 Aug 2009 00:15:46 +1000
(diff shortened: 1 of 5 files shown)


oz:
Hi there!

Is something blocking this PR? It's a small change but:

1. it would help avoiding monkey-patching in other projects, and
2. make Ruby's emails a tad less likely to end up in spam.

Please let me know if I could help if there's an identified blocker. 🙏🏻

c960657:
Fixed in #1503.

ukolovda:
I check latest Gem version 2.8.1, but it still have old version (Mime-Version instead of MIME-Version).
https://github.com/mikel/mail/blob/2.8.1/lib/mail/fields/mime_version_field.rb

Please publish new release with this fix, if possible.

PR #1247, opened 2018-07-06
remove has rdoc

ahorek:
```
NOTE: Gem::Specification#has_rdoc= is deprecated with no replacement. It will be removed on or after 2018-12-01.
Gem::Specification#has_rdoc= called from /mail/mail.gemspec:13.
```


jeremy:
Backported to 2-7-stable @ 904fc3299163771ca47d5cc5c2f540f6f828c2d6

PR #1249, opened 2018-07-28
Use String.blank?

KevinColemanInc:
I use https://github.com/SamSaffron/fast_blank . I would prefer if mail to use String.blank? if it is available.

### DIFF
--- a/lib/mail/utilities.rb
+++ b/lib/mail/utilities.rb
@@ -310,7 +310,7 @@ module Mail
       if value.kind_of?(NilClass)
         true
       elsif value.kind_of?(String)
-        value !~ /\S/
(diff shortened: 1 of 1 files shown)


KevinColemanInc:
@mikel Do you have any feedback for me about this PR?

lib/mail/utilities.rb
```diff
@@ -310,7 +310,7 @@ def blank?(value)
       if value.kind_of?(NilClass)
         true
       elsif value.kind_of?(String)
-        value !~ /\S/
+        value.respond_to?(:blank?) ? value.blank? : value !~ /\S/
```
jeremy: Would make this the first conditional check rather than making it String-specific

PR #878, opened 2015-05-16
Fix #844 some old mailer client use 'uuencode'

GaetanJUVIN:
Handle 'uuencode' by registering UnixToUnix (same as 'x-uuencode')

Issue #844, opened 2014-12-31
Please add support uuencode or explain how to download the file


yuri-karpovich:
I'm trying to download uuencode attachment:

```
   content = email.message.attachments[0].body
   File.open("#{dir}/#{file_name}" , "wb") { |file| file.write(content) }
```

=> Mail::UnknownEncodingType: Don't know how to decode **uuencode**, please call #encoded and decode it yourself.

Can you please explain how can I download it without errors?

bf4:
Please include some the failing code, along with your Ruby version, Mail gem version, OS, etc. per https://github.com/mikel/mail/blob/master/CONTRIBUTING.md

PR #1250, opened 2018-07-30
additionally register UnixToUnix encoding as 'x-uue'

jkraemer:
Processing an email that used 'x-uue' as encoding name failed. Problem and fix are similar to #844 / #878


jeremy:
Merged 8266b3df4fceeca65dfea13796411bdb906d5dba and backported to 2-7-stable 870fc4fbb90a70eda26e05609e07bf7cc666497d. Thank you for contributing, @jkraemer!

PR #1256, opened 2018-09-07
use accessors

ahorek:


ahorek:
I think attrs are more readable then methods + less code. They're also slighty faster.
Rdoc should be changed as well or I could split attr_accessors to attr_reader, attr_wríter and leave the original?

ahorek:
I've changed docs and do some reformatting. Is it ok?

PR #1258, opened 2018-09-07
default parser

ahorek:
URI::Parser.new always allocates a new instance, but we can use URI::DEFAULT_PARSER instead

### DIFF
--- a/lib/mail/utilities.rb
+++ b/lib/mail/utilities.rb
@@ -159,7 +159,7 @@ module Mail
     end
 
     def uri_parser
-      @uri_parser ||= URI.const_defined?(:Parser) ? URI::Parser.new : URI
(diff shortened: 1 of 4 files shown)


lib/mail/utilities.rb
```diff
@@ -159,7 +159,7 @@ def uri_unescape( str )
     end
 
     def uri_parser
-      @uri_parser ||= URI.const_defined?(:Parser) ? URI::Parser.new : URI
+      @uri_parser ||= URI.const_defined?(:Parser) ? URI::DEFAULT_PARSER : URI
```
jeremy: Should change the const_defined? check?

ahorek: Thanks, yes it should be changed

ahorek:
fixed

PR #1259, opened 2018-09-07
tr

ahorek:


ahorek:
I did minor changes to improve performance of mail based on rubocop and fast-ruby.
https://github.com/JuanitoFatas/fast-ruby/blob/master/README.md

Gsub is ment for more complex substitutions. If you want to change only a single character it can be subtituted by tr which is about 5 times faster.

lib/mail/fields/unstructured_field.rb
```diff
@@ -175,7 +175,7 @@ def encode(value)
       value.gsub!(/\)/, '=29')
       value.gsub!(/\?/, '=3F')
       value.gsub!(/_/,  '=5F')
-      value.gsub!(/ /,  '_')
+      value.tr!(' ',  '_')
```
jeremy: This one feels out of place alongside a series of `gsub!`. What's the performance impact / is it justified?

lib/mail/fields/unstructured_field.rb
```diff
@@ -186,7 +186,7 @@ def encode_crlf(value)
     end
 
     def normalized_encoding
-      encoding = charset.to_s.upcase.gsub('_', '-')
+      encoding = charset.to_s.upcase.tr('_', '-')
```
jeremy: Not a performance hotspot, but it reads well 👍

Issue #1269, opened 2018-09-20
mail crashes (unhanded exception) upon initializing a big date field


zelivans:
Hi, I've been testing mail with a Ruby fuzzing tool ([kisaten](https://github.com/zelivans/kisaten)) and found a few bugs. This is the first and simple bug I found.

The `tst.rb` file I used simply calls `Mail.read` with given input. The input that causes said crash:

## Input
```
Subject:0000=0000000
Content-Trans0@0000000000>
To:000@0000000000
Content-Dispcsition:000;000000000=0000000
Content-Transfer-Content-Dispositiontent-Description:0      0
Message-Id:<000000Content-Description000000000000000@000000000>
Msion:0.0
Date:200000000000/00A0000 0000=0000000
```

I left the input unchanged, but only the `Date` field is actually relevant.

## Raw crash
```
/home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/fields/common_date_field.rb:17:in `parse': integer 200000000000 too big to convert to `int' (RangeError)
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/fields/common_date_field.rb:17:in `normalize_datetime'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/fields/common_date_field.rb:31:in `initialize'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/fields/common_field.rb:12:in `new'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/fields/common_field.rb:12:in `parse'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/field.rb:267:in `parse_field'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/field.rb:256:in `create_field'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/field.rb:183:in `field'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/field.rb:229:in `method_missing'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/field_list.rb:30:in `add_field'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/header.rb:100:in `block in fields='
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/header.rb:98:in `each'
[...]

PR #1270, opened 2018-09-20
Prevent unhandled RangeError in CommonDateField

zelivans:
Fixes https://github.com/mikel/mail/issues/1269

The `normalize_datetime` method causes a `RangeError` crash when `DateTime.parse` fails to parse a date too big. This fixes this issue naively by using the original string when a `RangeError` is raised from the DateTime call.

### DIFF
--- a/lib/mail/fields/common_date_field.rb
+++ b/lib/mail/fields/common_date_field.rb
@@ -15,6 +15,8 @@ module Mail
         stripped = string.to_s.gsub(/\(.*?\)/, '').squeeze(' ')
         begin
           datetime = ::DateTime.parse(stripped)
+        rescue RangeError
+        # If fails to convert because date is out of range, fall back to use the string, no need to raise
         rescue ArgumentError => e
           raise unless 'invalid date' == e.message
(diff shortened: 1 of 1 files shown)


rosa:
I run into this as well! @zelivans what do you think of adding also a test for this? This is the one I had added myself before realising there was already a PR for this:
```diff
diff --git a/spec/mail/fields/date_field_spec.rb b/spec/mail/fields/date_field_spec.rb
index bdb9e916..720f4c7c 100644
--- a/spec/mail/fields/date_field_spec.rb
+++ b/spec/mail/fields/date_field_spec.rb
@@ -66,5 +66,10 @@ describe Mail::DateField do
       field = Mail::DateField.new("12 Aug 2009 30:00:02 GMT")
       expect(field.date_time).to be_nil
     end
+
+    it "should handle out of integer range errors in date" do
+      field = Mail::DateField.new("12 Aug 2009 12:00:8184237681594 GMT")
+      expect(field.date_time).to be_nil
+    end
   end
 end
```

PR #1272, opened 2018-09-21
Prevent unhandled NoMethodError when date_time in ReceivedField is invalid

### LINKED ISSUE #1271: mail crashes upon invalid Received date field
This is another bug I found by fuzzing mail.

## Raw crash
```
/home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/fields/received_field.rb:40:in `formatted_date': undefined method `strftime' for nil:NilClass (NoMethodError)
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/fields/received_field.rb:48:in `do_encode'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/fields/common_field.rb:59:in `encoded'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/field.rb:229:in `method_missing'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/header.rb:182:in `block in encoded'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/header.rb:181:in `each'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/header.rb:181:in `encoded'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/message.rb:1803:in `encoded'
	from /home/ariel/afl-kisaten/private/sandbox/mail/mail/lib/mail/message.rb:1868:in `to_s'
	from tst.rb:3:in `puts'
	from tst.rb:3:in `puts'
	from tst.rb:3:in `<main>'
```

## Example input
```
Return-Path:<0@0>
Received:000000000000 (000000000000)0000000000000000000 (0000000000000000000000000000000000);        Wed, 90 May00000 00:00:00 -0000 (000)
Message-Id:<0@0>
From:000 00000 <0000@00000>
To:0000000 00000 <0000@00000>
```

Only the `Received` field is relevant.

## Explanation

This crash is triggered when trying to get the encoded result of a mail with an invalid received date. Specifically the `ReceivedElement` puts `nil` in its `date_time` when it can't parse the date, and later `ReceivedField` calls `strftime` on this object without checking that it is not nil.

I propose a simple fix by checking that `strftime` can be called on the `date_time` and issuing a warning when mail fails to parse an invalid date.

zelivans:
Fixes #1271

### DIFF
--- a/lib/mail/elements/received_element.rb
+++ b/lib/mail/elements/received_element.rb
@@ -28,7 +28,7 @@ module Mail
         ::DateTime.parse("#{received.date} #{received.time}")
       rescue ArgumentError => e
         raise e unless e.message == 'invalid date'
-        warn "WARNING: Invalid date field for recieved element (#{received.date} #{received.time}): #{e.class}: #{e.message}"
+        warn "WARNING: Invalid date field for received element (#{received.date} #{received.time}): #{e.class}: #{e.message}"
         nil
       end
   end
--- a/lib/mail/elements/received_element.rb
+++ b/lib/mail/elements/received_element.rb
@@ -28,6 +28,7 @@ module Mail
         ::DateTime.parse("#{received.date} #{received.time}")
       rescue ArgumentError => e
         raise e unless e.message == 'invalid date'
+        warn "WARNING: Invalid date field for recieved element (#{received.date} #{received.time}): #{e.class}: #{e.message}"
         nil
       end
   end
--- a/lib/mail/fields/received_field.rb
+++ b/lib/mail/fields/received_field.rb
@@ -37,7 +37,9 @@ module Mail
     end
 
     def formatted_date
-      date_time.strftime("%a, %d %b %Y %H:%M:%S ") + date_time.zone.delete(':')
+      if date_time.respond_to? :strftime and date_time.respond_to? :zone
+        date_time.strftime("%a, %d %b %Y %H:%M:%S ") + date_time.zone.delete(':')
+      end
     end
 
     private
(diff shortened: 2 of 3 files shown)


lib/mail/elements/received_element.rb
```diff
@@ -28,6 +28,7 @@ def datetime_for(received)
         ::DateTime.parse("#{received.date} #{received.time}")
       rescue ArgumentError => e
         raise e unless e.message == 'invalid date'
+        warn "WARNING: Invalid date field for recieved element (#{received.date} #{received.time}): #{e.class}: #{e.message}"
```
jeremy: received*

zelivans: Oops, thanks for pointing that out. Resolved in ce905c2

lib/mail/fields/received_field.rb
```diff
@@ -37,7 +37,9 @@ def info
     end
 
     def formatted_date
-      date_time.strftime("%a, %d %b %Y %H:%M:%S ") + date_time.zone.delete(':')
+      if date_time.respond_to? :strftime and date_time.respond_to? :zone
+        date_time.strftime("%a, %d %b %Y %H:%M:%S ") + date_time.zone.delete(':')
+      end
```
jeremy: Let's add a test case that demonstrates that `#formatted_date` is `nil` for an invalid field 👍

zelivans: Turns out there was a test case for an invalid date in `received_field_spec.rb` but it did not check `#formatted_date`. I verified this was indeed failing on the master branch. Added the test in 3e11433

PR #1280, opened 2018-10-10
Refactoring flatten.map

rubonz:


lib/mail/message.rb
```diff
@@ -285,7 +285,7 @@ def reply(*args, &block)
           bracketed_message_id = "<#{message_id}>"
           reply.in_reply_to = bracketed_message_id
           if !references.nil?
-            refs = [references].flatten.map { |r| "<#{r}>" }
```
ahorek: `.flat_map { }` equals to `.map { }.flatten`

`.flatten.map { }` is something different, it can't be changed this way

lib/mail/message.rb
```diff
@@ -285,7 +285,7 @@ def reply(*args, &block)
...
+            refs = [references].flat_map { |r| "<#{r}>" }
```
jeremy: `Array#flat_map` is unavailable on Ruby 1.8, unfortunately.

We can use the `ReferencesField` directly, however. Check whether it's present in the message header then request its message_ids, which is always an Array:
```ruby
if references_field = headers[:references]
  reply.references = references_field.message_ids.map { … } + [ bracketed_message_id ]
elsif …
```

PR #1287, opened 2018-11-01
Allowing multipart messages to be split when they are separated by LF only instead of CRLF

ppadron:
Some multipart messages use \n only instead of \r\n in their boundaries. Ran into this while parsing feedback loop reports sent by Return Path.

### DIFF
--- a/lib/mail/body.rb
+++ b/lib/mail/body.rb
@@ -267,7 +267,7 @@ module Mail
       parts_regex = /
         (?:                    # non-capturing group
           \A                |  # start of string OR
-          \r\n                 # line break
+          \r?\n                # line break
          )
         (
           --#{Regexp.escape(boundary || "")}  # boundary delimiter


jeremy:
Thanks @ppadron! We convert line endings to \n before parsing which covers this case. Could you share an example of where it fails, including the raw email, and on which version of the Mail library?

ppadron:
Hi @jeremy! I created a repository with a proof of concept for this issue: https://github.com/ppadron/mail-poc-pr-1287. In the Gemfile you can switch the mail version and see that in 2.7.1 it detects only one part, and with the fix it splits the email in 3 parts.

sebbASF:
See #1319 which has the same fix, but also removes the conversion to CRLF (*).
It also adds a test case that fails without the change.

(*) The conversion does not happen anyway unless the mail string is ascii_only, which is not the case for the failing test cases.

jeremy:
Going ahead with the additional changes in #1319. Thanks again @ppadron!

PR #1288, opened 2018-11-02
Split long subject to avoid RFC2047 violation, but only when String#scan is aware of multibyte strings.

qnighy:
I've hit RFC2047 violation ("An 'encoded-word' may not be more than 75 characters long") when sending a mail with a long Japanese subject. This led to a corrupted subject in some mailers, at least in Windows 10 Mail in conjunction with GMail IMAP.

I found that in history of `mail`, there was a character boundary problem with splitting multibyte characters. Therefore, in this patch, it only tries to split encoded words when `String#scan` seems to be able to split multibyte strings appropriately.

It doesn't check the RFC2047 constraint directly, but I think in most cases the encoded words will be within 75 bytes.


qnighy:
Rebased.

Issue #1299, opened 2018-12-17
[master] missing charset regression in action mailer


ahorek:
bisected to https://github.com/mikel/mail/commit/404ecdd290f89658fb05da02ea5b6139fd82ebdb

backtrace
```
NoMethodError: undefined method `ascii_only?' for nil:NilClass
/mail/lib/mail/fields/parameter_hash.rb:47:in `block in encoded'
/mail/lib/mail/fields/parameter_hash.rb:46:in `map!'
/mail/lib/mail/fields/parameter_hash.rb:46:in `encoded'
/mail/lib/mail/fields/content_type_field.rb:97:in `encoded'
/mail/lib/mail/field.rb:229:in `method_missing'
/mail/lib/mail/header.rb:182:in `block in encoded'
/mail/lib/mail/header.rb:181:in `each'
/mail/lib/mail/header.rb:181:in `encoded'
/mail/lib/mail/message.rb:1803:in `encoded'
/mail/lib/mail/body.rb:147:in `block in encoded'
/mail/lib/mail/parts_list.rb:31:in `block in collect'
C:/Ruby25-x64/lib/ruby/2.5.0/delegate.rb:349:in `each'
C:/Ruby25-x64/lib/ruby/2.5.0/delegate.rb:349:in `block in delegating_block'
/mail/lib/mail/parts_list.rb:31:in `collect'
/mail/lib/mail/body.rb:147:in `encoded'
/mail/lib/mail/message.rb:1805:in `encoded'
C:/Ruby25-x64/lib/ruby/gems/2.5.0/gems/actionmailer-5.2.2/lib/action_mailer/base.rb:575:in `set_payload_for_mail'
C:/Ruby25-x64/lib/ruby/gems/2.5.0/gems/actionmailer-5.2.2/lib/action_mailer/base.rb:559:in `block in deliver_mail'
C:/Ruby25-x64/lib/ruby/gems/2.5.0/gems/activesupport-5.2.2/lib/active_support/notifications.rb:168:in `block in instrument'
C:/Ruby25-x64/lib/ruby/gems/2.5.0/gems/activesupport-5.2.2/lib/active_support/notifications/instrumenter.rb:23:in `instrument'
C:/Ruby25-x64/lib/ruby/gems/2.5.0/gems/activesupport-5.2.2/lib/active_support/notifications.rb:168:in `instrument'
[...]

PR #1300, opened 2018-12-18
empty charset fix

ahorek:
fixes https://github.com/mikel/mail/issues/1299

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -1477,7 +1477,9 @@ module Mail
           warning = "Non US-ASCII detected and no charset defined.\nDefaulting to UTF-8, set your own if this is incorrect.\n"
           warn(warning)
         end
-        header[:content_type].parameters['charset'] = @charset
+        if @charset
+          header[:content_type].parameters['charset'] = @charset
+        end
       end
     end
 
--- a/lib/mail/header.rb
+++ b/lib/mail/header.rb
@@ -169,7 +169,11 @@ module Mail
     def charset=(val)
       params = self[:content_type].parameters rescue nil
       if params
-        params[:charset] = val
+        if val
+          params[:charset] = val
+        else
+          params.delete(:charset)
+        end
       end
       @charset = val
     end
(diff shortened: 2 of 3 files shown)


ahorek:
the actual problem is in action mailer, that always expects a charset to be defined and https://github.com/mikel/mail/pull/1145 broke it, however I think it should be fixed here, otherwise it'll break older rails versions

@mikel what about add action mailer's tests to travis?

jeremy:
@ahorek, that's a good idea. Do Action Mailer tests fail with #1145? We could test Action Mailer against Mail master branch as well.

ahorek:
> Do Action Mailer tests fail with #1145?

yes, they do fail. I'm working on CI integration in #1303

jeremy:
Great! Let's get a PR going in Rails as well. We can track Mail master branch in an additional CI job there.

lib/mail/message.rb
```diff
@@ -1503,8 +1503,10 @@ def charset
     # Sets the charset to the supplied value.
     def charset=(value)
       @defaulted_charset = false
-      @charset = value
-      @header.charset = value
+      if value
+        @charset = value
+        @header.charset = value
+      end
```
jeremy: How would someone remove the charset now, to rely on default parsing?

This is seeming like something to fix downstream in Action Mailer.

ahorek: it isn't just "bad tests", it breaks Action Mailer for real. Even if it'll be fixed in Action Mailer, Mail 2.8 will be incompatible with all previous Rails versions, I think we should avoid that for all costs.

what about skipping blank parameters? https://github.com/mikel/mail/blob/master/lib/mail/fields/parameter_hash.rb#L47

ahorek: ok, I fixed that by removing the existing header if you set charset = nil

note that this case didn't work even on previous mail versions including 2.7.1 stable

ahorek:
if you have time, please review https://github.com/mikel/mail/pull/1308 https://github.com/mikel/mail/pull/1301 these two PRs are preventing a green build on travis

ahorek:
ping @jeremy 

it should be ready, what do you think? any objections? travis build is still failing because of https://github.com/mikel/mail/pull/1308

PR #1301, opened 2018-12-19
fix spec

ahorek:
address failing tests due to deprication in active support
https://github.com/rails/rails/blob/d684778a000cdcfcac53f6390c565c6a91a3391e/activesupport/lib/active_support/multibyte/chars.rb#L79


ahorek:
cc @jeremy is it the right way to fix the problem? I don't really remember how ruby 1.8 deals with encodings, but I hope it's correct. Tests are green again.

ahorek:
restarted build due to https://github.com/mikel/mail/pull/1306

PR #1302, opened 2018-12-19
get rid of rescue nil

ahorek:
not a good idea, nil check should be enough


jeremy:
Believe this is meant to catch parsing errors. Agree that blanket `rescue nil` is too broad, though. We can be more precise.

ahorek:
I tested this change with 3 test suites (mail, actionmailer and redmine) wihout any regression. These methods aren't responsible for parsing, it should be just a lookup. Rescuing StandardError is too cautious, but well, still better than rescue Exception 😈

ahorek:
rebased

PR #1303, opened 2018-12-19
[travis] - add actionmailer

ahorek:
add action mailer job https://github.com/mikel/mail/pull/1300

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,10 +33,17 @@ env:
     - MBCHARS=builtin
 
 matrix:
+  include:
+    - name: actionmailer
+      script: spec/travis-actionmailer.sh
+      rvm: 2.5
+      install: skip
+      env: BUNDLE_GEMFILE=$PWD/rails/Gemfile
   allow_failures:
     - rvm: ruby-head
     - rvm: jruby
     - rvm: jruby-head
     - rvm: rbx-2
     - rvm: rbx-3
+    - name: actionmailer
   fast_finish: true
--- a/spec/travis-actionmailer.sh
+++ b/spec/travis-actionmailer.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+set -v -e
+
+git clone --depth=1 https://github.com/rails/rails.git -b master
+cd rails
+echo -e "\ngem 'mail', path: '../'" >> Gemfile
+bundle update --source mail
+cd actionmailer
+bundle exec rake test


ahorek:
@jeremy do you have an idea how to fix? could you try to upgrade rvm and gpg keys manually?

https://travis-ci.org/mikel/mail/jobs/470182472
```
rvm get master
Downloading https://get.rvm.io
Downloading https://raw.githubusercontent.com/rvm/rvm/master/binscripts/rvm-installer.asc
Verifying /home/travis/.rvm/archives/rvm-installer.asc
gpg: Signature made Wed 12 Dec 2018 04:06:03 PM UTC using RSA key ID 39499BDB
gpg: Can't check signature: No public key
Warning, RVM 1.26.0 introduces signed releases and automated check of signatures when GPG software found. Assuming you trust Michal Papis import the mpapis public key (downloading the signatures).
GPG signature verification failed for '/home/travis/.rvm/archives/rvm-installer' - 'https://raw.githubusercontent.com/rvm/rvm/master/binscripts/rvm-installer.asc'! Try to install GPG v2 and then fetch the public key:
    gpg2 --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
or if it fails:
    command curl -sSL https://rvm.io/mpapis.asc | gpg2 --import -
the key can be compared with:
    https://rvm.io/mpapis.asc
    https://keybase.io/mpapis
NOTE: GPG version 2.1.17 have a bug which cause failures during fetching keys from remote server. Please downgrade or upgrade to newer version (if available) or use the second method described above.
/home/travis/.rvm/scripts/functions/cli: line 243: return: _ret: numeric argument required
```

.travis.yml
```diff
@@ -21,8 +21,7 @@ rvm:
   - ruby-head
   - jruby
   - jruby-head
-  - rbx-2
-  - rbx-3
+  - truffleruby
```
jeremy: Let's leave rbx as-is and add truffleruby in a separate PR. This PR can focus on integration testing with Action Mailer alone.

.travis.yml
```diff
@@ -33,10 +32,16 @@ env:
     - MBCHARS=builtin
 
 matrix:
+  include:
+    - name: actionmailer
+      script: spec/travis-actionmailer.sh
+      rvm: 2.5
+      install: skip
+      env: BUNDLE_GEMFILE=$PWD/rails/Gemfile
```
jeremy: Is `$PWD` needed here? Can we rely on the current working directory?

spec/travis-actionmailer.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+set -v -e
+
+rm -rf rails
+git clone --depth=1 https://github.com/rails/rails.git -b master
+cd rails
+echo "" >> Gemfile
+echo "gem 'mail', path: '../'" >> Gemfile
+rm -rf Gemfile.lock
+bundle install
```
jeremy: `bundle update --source mail` may be sufficient, rather than removing the lock and reinstalling

spec/travis-actionmailer.sh
```diff
@@ -0,0 +1,13 @@
...
+cd actionmailer
+bundle install
```
jeremy: Second bundle install is superfluous. Can remove it.

.travis.yml
ahorek: Travis sets the variable by default to $PWD/Gemfile, so I have to override it, otherwise bundle update in the rails repository will use the main Gemfile from the mail gem and it'll break the build

.travis.yml
ahorek: ok

PR #1308, opened 2019-01-09
bundler for older rubies

ahorek:
https://bundler.io/blog/2019/01/03/announcing-bundler-2.html

Removed support for Ruby < 2.3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ sudo: false
 cache: bundler
 
 bundler_args: --without local_development
-before_install:
-  - gem install bundler
 
 script: bundle exec rake
 


.travis.yml
```diff
@@ -1,10 +1,7 @@
 language: ruby
-sudo: false
```
jeremy: Removing this because it's now default, or is it related to this PR?

ahorek: it isn't necessary to fix tests but it's recommended to remove the line, see
https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

jeremy:
Cool. Let’s do it in a separate PR then.

ahorek:
@jeremy done

PR #1309, opened 2019-01-09
#join remove empty strings

ahorek:
.join('') allocates an empty string which isn't necessary especially on older rubies that don't support frozen string literals


ahorek:
rebased against the current master

Issue #1225, opened 2018-04-15
Difference between `deliver` and `deliver!` response


dv:
Using `deliver!` with `return_response: true` returns the SMTP response, however using `deliver` ignores `return_response` and always returns `self`.

Is there a reason for this difference? Returning the SMTP response is handy for example when sending through Amazon SES, which adjusts and returns the MessageID in the SMTP response.

If I make a PR to make this consistent, would it be accepted?

jeremy:
No reason that I could discern from https://github.com/mikel/mail/commit/15c3fa7ce152d4c1da7fb0caa391ebf847d6fcd7. Consistent behavior is welcome 👍

PR #1313, opened 2019-01-30
Have Mail.deliver method uphold settings[:return_response]

vkozyrev:
I have updated this method to have consistent behavior with deliver!

This follows up on issue https://github.com/mikel/mail/issues/1225

Does this look like something you would be willing to merge in? Or are there any issues that I missed.

Currently 4 tests failing after the update
```
rspec ./spec/mail/network_spec.rb:259 # Mail deliveries perform_deliveries should call deliver! on the delivery method by default
rspec ./spec/mail/network_spec.rb:266 # Mail deliveries perform_deliveries should not call deliver if perform deliveries is set to false
rspec ./spec/mail/network_spec.rb:317 # Mail deliveries raise_delivery_errors should not pass on delivery errors if raised raise_delivery_errors is set to false
rspec ./spec/mail/network_spec.rb:349 # Mail deliveries delivery_handler mail only call its delivery_method once
```

I can fix tests once I get some initial feedback

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -249,13 +249,13 @@ module Mail
     #  mail.deliver
     def deliver
       inform_interceptors
-      if delivery_handler
+      response = if delivery_handler
         delivery_handler.deliver_mail(self) { do_delivery }
       else
         do_delivery
       end
       inform_observers
-      self
+      delivery_method.settings[:return_response] ? response : self
     end
 
     # This method bypasses checking perform_deliveries and raise_delivery_errors,


lib/mail/message.rb
```diff
@@ -249,13 +249,14 @@ def inform_interceptors
...
-      if delivery_handler
+      response = if delivery_handler
         delivery_handler.deliver_mail(self) { do_delivery }
       else
         do_delivery
       end
       inform_observers
-      self
+      
+      response
```
vkozyrev: I just return response because it seems `delivery_method.settings[:return_response]` is checked in each specific `delivery_method.deliver!` implementation

PR #1316, opened 2019-02-12
Part spec

sebbASF:
Same fix applies to master ...


sebbASF:
Sorry, did not mean to include all those files in the PR

PR #1319, opened 2019-02-13

sebbASF:
Conversion of body to CRLF does not help here because body is not ascii_only
so does not get converted. Instead to search for CR?LF (optional CR)
in which case conversion to CRLF is unnecessary.


jeremy:
Thanks @sebbASF! Dig this.

akissinger:
I ran into this error myself, and nearly made the same PR (without the nice unit tests of course...). Is this going to get merged at some point?

mikel:
Waiting for specs to be working on the master forward port of this.

PR #1320, opened 2019-02-13
Span => Spam

sebbASF:
Fix typo and broken local link


sebbASF:
Note: same typo appears in 2-7-stable etc

Issue #659, opened 2014-02-01
SSL version option


AlexRevin:
Hello

As MS Outlook does not correctly work with default ssl context on both Mac and Debian

I suggest adding a _:ssl_version_ option, so that this hack could be avoided

``` ruby
context.ssl_version = :SSLv3
```

in `lib/mail/network/delivery_methods/smtp.rb`

https://github.com/AlexRevin/mail/commit/a12f77d49c3957bddce3b76cf67587f76fa1acc9#diff-e3729936fa9291d1a4ff988ac78acc12

jeremy:
Open to PRs to expose the SSL context for custom config. Workaround in the meantime: use the SMTP connection delivery method instead and pass your own SMTP connection that you've manually configured. Other workaround: modify `Net::SMTP.default_ssl_context` to suit.

PR #1333, opened 2019-04-09
Support setting of SSL version

stanhu:
Closes https://github.com/mikel/mail/issues/659

### DIFF
--- a/spec/mail/network/delivery_methods/smtp_spec.rb
+++ b/spec/mail/network/delivery_methods/smtp_spec.rb
@@ -210,16 +210,14 @@ describe "SMTP Delivery Method" do
     it "should set SSL version if one is given" do
       context = OpenSSL::SSL::SSLContext.new
       allow(Net::SMTP).to receive(:default_ssl_context).and_return(context)
-      expect(context).to receive(:min_version=).with(OpenSSL::SSL::TLS1_1_VERSION).at_least(1).and_call_original
-      expect(context).to receive(:max_version=).with(OpenSSL::SSL::TLS1_2_VERSION).at_least(1).and_call_original
+      expect(context).to receive(:ssl_version=).with(:TLSv1_2).at_least(1)
 
       Mail.defaults do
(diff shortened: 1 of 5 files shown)


jeremy:
Nice! Thanks @stanhu

spec/mail/network/delivery_methods/smtp_spec.rb
```diff
@@ -206,6 +206,28 @@ def redefine_verify_none(new_value)
 
       expect { mail.deliver! }.not_to raise_error
     end
+
+    it "should set verify mode if one is given" do
```
stanhu: Oops, typo here.

ye:
Hi, is there any updates on this PR? Is it ready to be merged?

russellsch:
Would be great if this would get merged! This would sort out some issues I've been having.

Brennwert:
This would also solve my problems. I'd support a merge, please.

kyrofa:
I'd like to see this merged and available as well, please.

twk3:
@jeremy it would be great to get this reviewed/merged. There are more and more servers requiring new tls versions.

stanhu:
It looks like the `ssl_version` is now deprecated. From https://docs.ruby-lang.org/en/2.7.0/OpenSSL/SSL/SSLContext.html:

> Sets the SSL/TLS protocol version for the context. This forces connections to use only the specified protocol version. This is deprecated and only provided for backwards compatibility. Use min_version= and max_version= instead.

timkleier:
@stanhu I noticed your commit to support `min_version` and `max_version`, is there a reason why it was then subsequently reverted? I'd love to see this supported

stanhu:
@timkleier Older Ruby versions use `ssl_version`, while Ruby 2.7+ uses `min_version` and `max_version`. We could probably set this dynamically, but this pull request hasn't received any feedback from the maintainer for years. I'm closing it because of that.

PR #1339, opened 2019-05-06
Fix invalid byte sequence error

jsteiner:
This fixes an error where decoding a UTF-7 string raised an
`Encoding::InvalidByteSequenceError`. The resulting string isn't
perfect. There's some weird characters, but the result is better than
the alternative.


jeremy:
Thanks @jsteiner !

PR #1341, opened 2019-05-27
Failing reference parsing

ylecuyer:
Hi,

We have noticed lately a parse error on the Reference header.
We never had trouble until we reached this one, but when I looked inside our db this issue is present for a lot of emails.

I have attached to this PR two test with the parsing failure. One is when references are separated with new line and the other is when the references are separated with comma

### DIFF
--- a/spec/mail/fields/references_field_spec.rb
+++ b/spec/mail/fields/references_field_spec.rb
@@ -1,7 +1,7 @@
 # encoding: utf-8
 # frozen_string_literal: true
 require 'spec_helper'
-# 
+#
 #    The "References:" field will contain the contents of the parent's
 #    "References:" field (if any) followed by the contents of the parent's
 #    "Message-ID:" field (if any).  If the parent message does not contain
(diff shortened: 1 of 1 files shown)


spec/mail/fields/references_field_spec.rb
```diff
@@ -26,6 +26,24 @@
...
+    t = Mail::ReferencesField.new('<1234@test.lindsaar.net>\r\n <5678@test.lindsaar.net>')
+    expect(t.name).to eq 'References'
+    expect(t.value).to eq '<1234@test.lindsaar.net> <5678@test.lindsaar.net>'
+    expect(t.message_id).to eq '1234@test.lindsaar.net'
+    expect(t.message_ids).to eq ['1234@test.lindsaar.net', '5678@test.lindsaar.net']
+    expect(t.to_s).to eq '<1234@test.lindsaar.net> <5678@test.lindsaar.net>'
+  end
+
+  it "should accept multiple message ids separated by comma" do
+    t = Mail::ReferencesField.new('<1234@test.lindsaar.net>,<5678@test.lindsaar.net>')
```
jeremy:
```ruby
>> Mail::ReferencesField.new("<1234@test.lindsaar.net>,<5678@test.lindsaar.net>").message_ids
=> ["1234@test.lindsaar.net", "5678@test.lindsaar.net"]
```

spec/mail/fields/references_field_spec.rb
```diff
@@ -26,6 +26,24 @@
     expect(t.message_id).to eq '1234@test.lindsaar.net'
   end
 
+  it "should accept multiple message ids separated by new line" do
+    t = Mail::ReferencesField.new('<1234@test.lindsaar.net>\r\n <5678@test.lindsaar.net>')
```
jeremy: Use double quoted strings so \r\n are treated as carriage return and newline:
```ruby
>> Mail::ReferencesField.new("<1234@test.lindsaar.net>\r\n <5678@test.lindsaar.net>").message_ids
=> ["1234@test.lindsaar.net", "5678@test.lindsaar.net"]
```

ylecuyer:
@jeremy I'm sorry I made you loose time yesterday. I brainfucked and now I understand why.
We couldn't parse this kind of References header because we use 2.7.1 and not master.

I'm trying to backport https://github.com/mikel/mail/commit/a83fcfaeab83b7ffbb635d17887734be1633ab58 into the 2.7.2 version to be released but this is not straightforward.

I tried to update to 2.8 using master branch but it is impossible because of this bug : https://github.com/mikel/mail/issues/1299 which seems to have been fixed here https://github.com/mikel/mail/pull/1300 but not merged

Eventually I think I'll try to upgrade using the mater branch using the commit just before https://github.com/mikel/mail/commit/404ecdd290f89658fb05da02ea5b6139fd82ebdb this should be the easiest way to have this all working

PR #1345, opened 2019-06-04
fix actionmailer spec

ahorek:
fixes

   /home/travis/.rvm/rubies/ruby-2.5.5/lib/ruby/2.5.0/rubygems.rb:284:in `find_spec_for_exe': Could not find 'bundler' (1.17.3) required by your /home/travis/build/mikel/mail/rails/Gemfile.lock. (Gem::GemNotFoundException)
   To update to the latest version installed on your system, run `bundle update --bundler`.
   To install the missing version, run `gem install bundler:1.17.3`

### DIFF
--- a/spec/travis-actionmailer.sh
+++ b/spec/travis-actionmailer.sh
@@ -4,6 +4,7 @@ set -v -e
 git clone --depth=1 https://github.com/rails/rails.git -b master
 cd rails
 echo -e "\ngem 'mail', path: '../'" >> Gemfile
-bundle update --source mail
+gem update bundler
+bundle update --bundler --source mail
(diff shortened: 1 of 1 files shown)


jeremy:
Thanks @ahorek! Seems we should install the missing version (1.17.3) rather than the system version?

ahorek:
https://github.com/rails/rails/blob/39d3c9bcacb43b0a45bcf06aa5b98f6e54deeb24/Gemfile.lock#L612

There's a lock for a specific version of Bundler in Rails. That is why a simple bundle update fails if the system version is different.

jeremy:
Ah I see. OK

PR #1347, opened 2019-06-17
Add init_with method to Mail::Body

lesleh:
Fixes the case where a `PartsList` gets serialized as an `Array`, but does not get correctly deserialized back to a `PartsList` instance. The `init_from` would ordinarily be added to the `PartsList` class but this does not work due to the use of `SerializeClass(Array)` so instead I assign it on deserialization of `Mail::Body`.

CC @jeremy

Related to https://github.com/mikel/mail/issues/1038

### DIFF
--- a/lib/mail/body.rb
+++ b/lib/mail/body.rb
@@ -50,6 +50,11 @@ module Mail
       set_charset
     end
 
+    def init_with(coder)
+      coder.map.each { |k, v| instance_variable_set(:"@#{k}", v) }
+      @parts = Mail::PartsList.new(coder['parts'])
+    end
+
     # Matches this body with another body.  Also matches the decoded value of this
     # body with a string.
     # 
--- a/spec/mail/body_spec.rb
+++ b/spec/mail/body_spec.rb
@@ -457,4 +457,12 @@ describe Mail::Body do
       expect(body.encoded).to eq 'The Body'
     end
   end
+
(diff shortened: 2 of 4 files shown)


lib/mail/body.rb
```diff
@@ -50,6 +50,11 @@ def initialize(string = '')
       set_charset
     end
 
+    def init_with(coder)
+      coder.map.each { |k, v| instance_variable_set(:"@#{k}", v) }
```
jeremy: Can we call `super` to pull in this implementation, or do we need to implement it ourselves?

lib/mail/body.rb
```diff
@@ -50,6 +50,11 @@ def initialize(string = '')
...
+      @parts = Mail::PartsList.new(coder['parts'])
```
jeremy: Nice fix; retains backward compatibility.

lib/mail/body.rb
lesleh: There's no implementation of it in the super class to call so I left it out.

lesleh:
@jeremy I added the test cases from that PR, and added one to ensure that `Mail::Message#parts` remains a `Mail::PartList` through serialisation and deserialisation.

jeremy:
Backported to 2-7-stable @ 3204c0b4733166b9664a552006286227dea09953

Issue #1329, opened 2019-03-20
Attachments/parts not parsed from real-world email


steobrien:
Here’s an anonymized version of an email we came across (a Jira notification):
<details>
 <summary>Raw email</summary>

```
Delivered-To: examplep@company.com
Received: by 2002:a05:6000:108d:0:0:0:0 with SMTP id y13csp13744073wrw;
        Wed, 13 Mar 2019 14:50:05 -0700 (PDT)
X-Google-Smtp-Source: APXvYqy7mPKcRctqXcBKcg5CVl4NueHCPf3g06cgi9tpGswLpM7zNJcetyO8G0irZe8DHF/gqJFo
X-Received: by 2002:ac8:7606:: with SMTP id t6mr16598956qtq.243.1552513805290;
        Wed, 13 Mar 2019 14:50:05 -0700 (PDT)
ARC-Seal: i=1; a=rsa-sha256; t=1552513805; cv=none;
        d=google.com; s=arc-20160816;
        b=moZWmZHtDWtOYtKPhoXN36zWN/TS1dpQqsyDQlrNX22q79KeA2OVUQJSfoKOT/LWIH
         b3kXF6IvIVpAjq0LYvagbl+u6YX79rDQDeC2J2eoioRB5LcpihZDyEy/68LzSUy89BsZ
         +ZhVlKTvqMV68qS9iLddoOg9NxdXStiOQ1e/ZAqtk6dLA7SrGjLSVt/vfWNoZFQ1YGfd
         Z/5Ccbq1qX0ZF8zVssEHVeRZoKbfPI5lcu2w633TagujbD8rN64XyfNs62bCtH9ET+SH
         ynZpJMobFRzGJ+4i5B9tS+HvmeMP9JUuDfQmFwgyiP6DiSxvLWAi14fvi+JqbnBsoPDn
         S2Bg==
ARC-Message-Signature: i=1; a=rsa-sha256; c=relaxed/relaxed; d=google.com; s=arc-20160816;
        h=feedback-id:precedence:auto-submitted:mime-version:subject
         :message-id:to:from:date:dkim-signature:dkim-signature;
        bh=DBHSNv0eq5QWpQ6X0SJlqwYKN6RZlnR94Fb5PlhzeEs=;
        b=niQS1cFFh0nWd3p/QtX98fF1P8lIhl8AZju3uKnUb6YO1CV6C7781pA37noEm1l7Nq
[...]

PR #1350, opened 2019-07-08
Handle edge case when inline image attachment part is missing a filename

lao9:
Fixes https://github.com/mikel/mail/issues/1329

First attempt at a fix: https://github.com/mikel/mail/pull/1332

In this second attempt, I tried to scope down the issue to where the issue appears most probable: inline images.

We do not add a default filename for the nameless attachment, rather just ensure that the message part is considered an attachment if it is an inline image.


lao9:
Hey @jeremy . If you have a free moment, I looked at your comments in #1332 and am proposing this fix instead. Let me know if you have any feedback.

steobrien:
I'd love to see this merged in soon to fix an issue we’re seeing. @jeremy, anything you're waiting on before accepting this patch?

lib/mail/message.rb
```diff
@@ -2135,6 +2135,14 @@ def find_attachment
       filename
     end
 
+    def inline_image?
+     image? && inline?
+    end
```
jeremy: `inline?` is a method on the Mail::Part subclass, so this'll break for a Mail::Message itself.

Could move this into Mail::Part and override the `attachment?` method, perhaps?

```ruby
# class Part < Message

def attachment?
  super || inline_image?
end

# …

def inline_image?
  # …
```

jeremy:
(Moved to 2.8.0 release as this technically breaks compatibility with current stable 2.7.x.)

lib/mail/message.rb
lao9: Thanks for the feedback - I'll try to get this in next by next week!

mikel:
@lao9 Please update this to latest master.

Moved to 2.9 as it breaks compatibility with 2.7 and 2.8.

PR #1360, opened 2019-09-07
SMTP settings via url

eval:
This is a rough sketch that allows the SMTP-settings to be configured via a url, e.g. `smtp(s)://user:pw@mailer.org?domain=sender.org`.

This makes applications easily configurable via `ENV['SMTP_URL']`.

Would this be something you'd be interested in?


eval:
Draft-status removed - feedback welcome!

mikel:
Nice @eval !  - Please rebase this onto latest master, include a changelog and I'll include it in 2.9

eval:
Hi @mikel, I updated the PR. I made the following changes:
- url-attributes now take precedence, so it's similar to how one would [configure the database in Rails](https://guides.rubyonrails.org/configuring.html#configuring-a-database).
- I used different default-settings in case of `smtp` for localhost and non-localhost.  
  This to make targeting a test-setup (e.g. mailcatcher) convenient, but also to require starttls by default for non-localhosts.

I tested the setup using mailcatcher and fastmail.

eval:
Updated docs&code to bring in line with the new acceptable values for `enable_starttls`.

eval:
@radar I rebased this one against master - maybe a candidate to include in v2.9?

PR #1364, opened 2019-11-10
Capitalization of MIME-Header

1resu:
The term "MIME" needs to capitalized as in the [RFC](https://tools.ietf.org/html/rfc2045) stated. A header in your notation will lead to a higher spam score in Rspamd: `MV_CASE (0.5)`


eugeneius:
This is a duplicate of https://github.com/mikel/mail/pull/1242, which also updates the tests and documentation.

jeremy:
Thanks @1resu!

PR #1371, opened 2019-12-26
Improve documentation

fwolfst:
- fix a copy/paste error in Mail#last
- **prior to conversation:** fix #decode example in README. Here I am not 100% sure but calling #decode on a mail-object gives me `Can not decode an entire message, try calling #decoded on the various fields and body or parts if it is a multipart message.`. Although I'd find it pretty cool if #decode gives me whatever text/body-part there is.
- **after conversation:** Slightly improved message of NoMethodError in `Message#decode`. Agreed on further improvements should include choosing a more distinct Error class.

### DIFF
--- a/lib/mail/mail.rb
+++ b/lib/mail/mail.rb
@@ -152,7 +152,7 @@ module Mail
     retriever_method.first(*args, &block)
   end
 
-  # Receive the first email(s) from the default retriever
+  # Receive the last email(s) from the default retriever
   # See Mail::Retriever for a complete documentation.
   def self.last(*args, &block)
     retriever_method.last(*args, &block)
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -1892,7 +1892,7 @@ module Mail
       when !self.multipart?
         body.decoded
       else
-        raise NoMethodError, 'Can not decode an entire message, try calling #decoded on the various fields and body or parts if it is a multipart message.'
+        raise NoMethodError, 'This message cannot be decoded as _entire_ message, try calling #decoded on the various fields and body or parts if it is a multipart message.'
       end
     end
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -349,7 +349,7 @@ mail.cc              #=> 'sam@test.lindsaar.net'
 mail.subject         #=> "This is the subject"
 mail.date.to_s       #=> '21 Nov 1997 09:55:06 -0600'
 mail.message_id      #=> '<4D6AA7EB.6490534@xxx.xxx>'
-mail.decoded         #=> 'This is the body of the email...
+mail.body.decoded    #=> 'This is the body of the email...
```
jeremy: `mail.decoded` is the way to go. It checks whether the message is plain text, an attachment, or non-multipart, and raises the exception you're seeing on multipart messages:
https://github.com/mikel/mail/blob/8471b01760196c7b510ddabce9c50f72c5e79a2f/lib/mail/message.rb#L1886-L1897

If you want to extract the text part of an email, if present:
```ruby
if mail.multipart?
  mail.text_part&.decoded
else
  mail.decoded
end
```

lib/mail/mail.rb
```diff
@@ -152,7 +152,7 @@ def self.first(*args, &block)
     retriever_method.first(*args, &block)
   end
 
-  # Receive the first email(s) from the default retriever
+  # Receive the last email(s) from the default retriever
```
jeremy: 👍🏼

README.md
fwolfst: Thanks for the explanation. If I'd had the time I would craft a PR to change the Exception (I probably did not look up what is happening because I saw the `NoMethodError`, alternatively the Exceptions message could be improved).

jeremy: Yeah, NoMethodError is inappropriate for this!

fwolfst: Updated PR with just slightly improved error message. Changing the NoMethodError (in `Undecodable` ?) should be a TODO (that I will not tackle).

mikel:
Thank you @fwolfst !

PR #1372, opened 2019-12-28
Only use dot-stuff workaround on old Ruby versions

c960657:
In #683 a workaround was introduced for [broken dot-stuffing in Net::SMTP](https://bugs.ruby-lang.org/issues/9627) in older Ruby versions. However, the workaround is applied to all versions of Ruby, causing additional leading dots in the email being sent.

Also, this PR fixes a bug in the original workaround. We only need to do manual dot-stuffing if the last line of the email contains only a dot with no trailing newline. The tests imply that we also need to dot-stuff all lines with single dots, but that is wrong and also causes additional dots in the email being sent.

## Steps to reproduce
```
Mail.new do |m|
  m.from "from@example.com"
  m.to   "to@example.com"
  m.body ".foo\n."
end.deliver!
```

### Expected body (as displayed in email client)
```
.foo
.
```

### Actual body
```
.foo
..
```

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -17,5 +17,7 @@ Compatibility:
 Features:
 * Message#inspect_structure and PartsList#inspect_structure pretty-print the hierarchy of message parts. (TylerRick)
 
+Bugs:
+* Fix duplicate dots at beginning of lines. (c960657)
 
 Please check [2-7-stable](https://github.com/mikel/mail/blob/2-7-stable/CHANGELOG.rdoc) for previous changes.
--- a/lib/mail/network/delivery_methods/smtp_connection.rb
+++ b/lib/mail/network/delivery_methods/smtp_connection.rb
@@ -66,11 +66,7 @@ module Mail
       end
 
       def dot_stuff(message)
-        if message.end_with?("\r\n.")
-          message << "."
-        end
-
-        message
+        message.gsub(/(\r\n\.)([^\r\n]*$)/, '\1.\2')
       end
   end
 end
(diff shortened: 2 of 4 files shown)


c960657:
Tested with Ruby 1.9.3, 2.0.0-p481, 2.3.3 and 2.7.0.

dalibor:
@jeremy Can you please release a new mail gem version with this fix? It is an obscure bug that people are rarely finding and it is affecting few Ruby versions plus the original regex was not right causing even more issues. Thanks!

dalibor:
Hm, this issue/fix explains the additional dot for a line with a single dot and is probably not affecting as many as I thought. 

There is another issue that I'm trying to figure out that is a missing dot. Any chance you have seen it?

I believe it happens with some buggy SMTP clients (or potentially antivirus programs or POP and IMAP in combination with) that most-likely removes the dot twice. This is very infrequent, but unfortunately it is happening.

I read about a [workaround](https://wordtothewise.com/2018/11/why-do-my-urls-have-two-dots/) that suggests encoding `.` at beginning of each line with `=2E` that will prevent the issue. I've seen [some](https://github.com/mailgun/flanker/pull/233/files) [libraries](https://github.com/swiftmailer/swiftmailer/blob/ff832c46efb93bd8c7ec86273fc292fc170b76a1/lib/classes/Swift/Mime/ContentEncoder/QpContentEncoder.php#L47) do that. Do you think we could extend `Mail::Encodings::QuotedPrintable.encode` with such a patch?

Also, I wonder why is Ruby's quoted-printable format 74 chars line length vs others splitting the lines at the max of 76 chars? That actually could help us by simplifying the fix to replace the first `.` with `=2E`, so with -1 + 3 we'll end-up with lines with max 76 chars so we don't have to split the line, but I'm not sure if we can rely on that the length is 74 chars?

@c960657 @jeremy Do you maybe have any thoughts about this?

PR #1375, opened 2020-01-20
Handle incomptible encodings in email headers

bobjflong:
Previously, running this new test case would throw an `Encoding::CompatibilityError` error, as it attempts to join the parsed ASCII-8BIT string and parsed UTF-8 string together.

Now, if a set of incompatible encodings is detected, we first force the strings to be UTF-8, and replace invalid characters.

Let me know if there are other thoughts on how to handle this error.


sakirtemel:
any updates here? and what should be the correct behaviour? I'm running similar issue with message-id header

mikel:
@bobjflong thanks for this. Wondering what real world case would exist where you are inserting incompatible strings?  Would we get this from an incoming email or is this only really possible when you are using mail in a program you control?

mikel:
cc @sakirtemel

sakirtemel:
@mikel as far as I remember it was happening when we were receiving an incoming email for our case. I've seen many of those happening and they were failing in our queues.

PR #1378, opened 2020-02-13
Use RSpec output matcher and do not redefine Kernel#warn in specs

eregon:
#1105 worked around by redefining `Kernel#warn` but that definition is incompatible with recent `Kernel#warn` which accepts keyword arguments (`:uplevel`), and of course it's best to not redefine core methods.
For example, an issue with that redefinition is https://github.com/oracle/truffleruby/issues/1900.

The RSpec [`output` matcher](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/output-matcher) makes it possible to be agnostic to how warnings are emitted, and just assert the output on `$stderr`.

cc @jeremyevans

### DIFF
--- a/spec/mail/field_spec.rb
+++ b/spec/mail/field_spec.rb
@@ -6,8 +6,11 @@ describe Mail::Field do
 
   describe 'parsing' do
     it "parses full header fields" do
-      expect($stderr).to_not receive(:puts)
-      field = Mail::Field.parse('To: Mikel')
+      field = nil
+      expect {
+        field = Mail::Field.parse('To: Mikel')
+      }.to_not output.to_stderr
+
       expect(field.name).to eq 'To'
       expect(field.value).to eq 'Mikel'
       if field.value.respond_to?(:encoding)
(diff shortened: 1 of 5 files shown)


jeremyevans:
I agree the specs shouldn't be overriding `Kernel#warn` and it was a mistake to do so.  I was unfortunately not familiar enough with RSpec to know how to test for warnings properly when I added the `Kernel#warn` override.  This pull request is a much better approach.

eregon:
@jeremyevans No worry, I think it was a fair enough approach that worked well enough back then.

spec/mail/field_spec.rb
```diff
@@ -6,8 +6,11 @@
 
   describe 'parsing' do
     it "parses full header fields" do
-      expect($stderr).to_not receive(:puts)
-      field = Mail::Field.parse('To: Mikel')
+      field = nil
+      expect {
+        field = Mail::Field.parse('To: Mikel')
+      }.to_not output.to_stderr
+
```
jeremy: TIL

PR #1379, opened 2020-02-14
Travis: attempt to fix builds on older Rubies

jeremy:
By running on Precise instead of Trusty. References travis-ci/travis-ci#4984.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ bundler_args: --without local_development
 script: bundle exec rake
 
 rvm:
-  - 1.8.7
-  - 1.9.2
   - 1.9.3
   - 2.0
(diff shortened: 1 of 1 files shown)


pirj:
Check this out [travis conf](https://github.com/rspec/rspec-core/blob/master/.travis.yml#L12), [`bundler` installer](https://github.com/rspec/rspec-core/blob/master/script/update_rubygems_and_install_bundler#L14), [Ruby version detection](https://github.com/rspec/rspec-core/blob/12caa2907b7eb67d65e19353c15613586858546c/script/predicate_functions.sh#L60).
This might at least fix the problem with Bundler 2.0 complaining about old Ruby version.

PR #1399, opened 2020-06-24
Remove downcase! method from sanitize function in ContentTypeField

### LINKED ISSUE #1398: sanitize function in ContentTypeField breaks encoding by using downcase!
I have an e-mail with PDF attachment, with the following `Content-Type` header:

```
Content-Type: application/pdf; name==?utf-8?B?RXhhbXBsZS5wZGY=?=
```

```
% echo "RXhhbXBsZS5wZGY=" | base64 -d
Example.pdf%
```

This header cannot be parsed using the `Mail::ContentTypeElement` because the encoded attachment name is not quoted. Inside `Mail::ContentTypeField`, there is a `sanitize` method which tries to work around these broken e-mails by performing actions, such as adding these quotes. However, this function also calls `downcase!` on the value:

https://github.com/mikel/mail/blob/8fbb17d4d5364c77cc870769d451bc2739b3a8ce/lib/mail/fields/content_type_field.rb#L126-L130

This results in the final header being:
```
Content-Type: application/pdf; name==?utf-8?b?rxhhbxbszs5wzgy=?=
```

Now the encoding is broken:
```
% echo "rxhhbxbszs5wzgy=" | base64 -d
?ao???p?
        %
```

Without the `downcase!`, the sanitize function would have fixed the problem just fine. With sanitize however, the encoding is broken.

RemcodM:
Fixes #1398

This change removes the sequence of code that downcases the Content-Type header value. If I should believe the specs, it should serve no purpose. Only 2 specs break because of the charset being returned in capital letters instead of downcased as before.

This fixes the problem is which encoding is broken by downcasing, such as the example with base64 provided in #1398

### DIFF
--- a/lib/mail/fields/content_type_field.rb
+++ b/lib/mail/fields/content_type_field.rb
@@ -123,12 +123,6 @@ module Mail
         gsub(/[; ]+/, '; '). #use '; ' as a separator (or EOL)
         gsub(/;\s*$/,'') #remove trailing to keep examples below
 
-      if val =~ /(boundary=(\S*))/i
-        val = "#{$`.downcase}boundary=#{$2}#{$'.downcase}"
-      else
-        val.downcase!
-      end
-
       case
       when val.chomp =~ /^\s*([\w\-]+)\/([\w\-]+)\s*;\s?(ISO[\w\-]+)$/i
         # Microsoft helper:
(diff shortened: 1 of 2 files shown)


ahorek:
nice find! I think you're right, the code should be removed. Could you add a test case according to your example to avoid further regressions?

mikel:
Hi there @RemcodM could you please rebase this onto latest master?  This will then trigger all the current specs we have.

PR #1403, opened 2020-08-15
Fix case of MIME-Version header field name

A1bi:
This changes the header field name `Mime-Version` to `MIME-Version`.

While an incorrect case here might not cause any trouble since header field names should be treated case-insensitively, the actual correct spelling of this header field is `MIME-Version` (see [RFC 2045](https://tools.ietf.org/html/rfc2045#section-4)).

More importantly I noticed that spam filters such as [Rspamd](https://rspamd.com) penalize emails with an incorrect case for header fields like this, see the screenshot below.

rspamd


mikel:
Thanks, merged this in https://github.com/mikel/mail/pull/1503 - I didn't see this PR :(

PR #1405, opened 2020-08-15
When parsing UTF-7, do not raise exceptions (except for B encodings)

### LINKED ISSUE #1404: Invalid utf-7 data in body causes exceptions
The following results in an `Encoding::InvalidByteSequenceError`:

```ruby
Mail::Part.new("Content-Type: text/plain; charset=\"utf-7\"\r\n\r\n&5g-").decoded
```

The equivalent using quoted-printable makes use of the replacement character:

```ruby
irb(main):005:0> Mail::Part.new("Content-Type: text/plain; charset=\"utf-8\"\r\n" +
                                "Content-Transfer-Encoding: quoted-printable\r\n\r\n=E6").decoded
=> "�"
```

There are [explicit tests](https://github.com/mikel/mail/blob/master/spec/mail/encodings_spec.rb#L215) that in B encodings we capture encoding errors and revert to displaying the raw text, but there's nothing trapping such errors when decoding the message body.

Should a utf-7 body work like quoted-printable or like a utf-7 header? In any case, it seems like a bad idea to let encoding-related exceptions bubble up to the application.

pdg137:
Fixes #1404 by removing UTF-7 encoding exceptions, calling `transcode_to_scrubbed_utf8` like most of the other code, without changing the behavior of B encodings.

Note: it's kind of ugly to have two versions of such an unusual decoding method, and I'm not sure that the kind of UTF-7 encoding supported by the Mail gem is valid for email bodies, but this at least should allow email to be processed without raising exceptions.

### DIFF
--- a/lib/mail/version_specific/ruby_1_9.rb
+++ b/lib/mail/version_specific/ruby_1_9.rb
@@ -15,10 +15,10 @@ module Mail
     end
 
     class BestEffortCharsetEncoder
-      def encode(string, charset)
+      def encode(string, charset, raise_utf7_exceptions = false)
         case charset
         when /utf-?7/i
-          Mail::Ruby19.decode_utf7(string)
+          Mail::Ruby19.decode_utf7(string, raise_utf7_exceptions)
         else
           string.force_encoding(pick_encoding(charset))
         end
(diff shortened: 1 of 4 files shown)


mikel:
Hi there @pdg137, first of all, thank you so much for this! We've dropped the `lib/mail/version_specific/ruby_1_9.rb` file now, could you please rebase these changes onto latest master and I'll look to merge then (if indeed still needed)

pdg137:
Sure! I can get to that in a day or two. But do you have any thoughts about just abandoning UTF-7 support for email bodies? I can look into it some more but expect it's only used for spam and even then just for a tiny fraction of it.

pdg137:
Okay, how does this look?

pdg137:
Could you remove the "More work required" label or let me know what else you are looking for?

pdg137:
Please let me know if you have more comments on this PR.

pdg137:
I looked into this a little more and made https://github.com/mikel/mail/issues/1618.

pdg137:
Again I'd like to know if there's anything blocking this from getting merged. It would be nice to get rid of this place that the Mail gem crashes.

PR #1409, opened 2020-09-09
Do not alter "name" field of Content-Type header

nbianca:
When Content-Type header parser fails, it tries to sanitize the value
and then parse it again. One step of the sanitization process was to
lowercase its value, which could change the meaning of "name" and
"filename" fields as upper and lower case characters represent
different things.

### DIFF
--- a/lib/mail/fields/content_type_field.rb
+++ b/lib/mail/fields/content_type_field.rb
@@ -125,6 +125,10 @@ module Mail
 
       if val =~ /(boundary=(\S*))/i
         val = "#{$`.downcase}boundary=#{$2}#{$'.downcase}"
+      elsif val =~ /(name=(\S*))/i
+        val = "#{$`.downcase}name=#{$2}#{$'.downcase}"
+      elsif val =~ /(filename=(\S*))/i
+        val = "#{$`.downcase}filename=#{$2}#{$'.downcase}"
       else
         val.downcase!
       end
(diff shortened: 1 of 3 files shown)


lib/mail/fields/content_type_field.rb
```diff
@@ -125,6 +125,10 @@ def sanitize(val)
 
       if val =~ /(boundary=(\S*))/i
         val = "#{$`.downcase}boundary=#{$2}#{$'.downcase}"
+      elsif val =~ /(name=(\S*))/i
```
SamSaffron, line 128: we are repeating this pattern 3 times now, does it make sense to just match on `([a-z]+)` and make the original clause more generic?

nbianca: I went for the minimum change to avoid breaking anything and I am under the belief that a smaller change might be merged faster. I do not honestly understand why the `.downcase` is required at all and if it were up to me, I'd remove the whole `if` block. I hope a maintainer who knows the gem better will clarify things here.

I am thinking that there is some lookup at some point (maybe for `charset`?) which is case-sensitive and that is why this has been added in the first place, but I believe that is the exception rather than the norm.

nbianca:
Hey @mikel and @jeremy,

Can you please review this?

lib/mail/fields/content_type_field.rb
nbianca: I did not go with the generic suggestion because I am not sure if that will not break something else, but I merged the three regular expressions, so the original clause is now more generic.

jeremy (approved):
Great diagnosis and fix. Thanks @nbianca!

PR #1425, opened 2020-12-21
Rubydoc link is broken

harirajv:
The rubydoc link in README is not pointing to the correct page.


mikel:
Thank you, but https://www.rubydoc.info/gems/mail seems to resolve correctly for me.

PR #1427, opened 2021-02-03
Add default timeouts for SMTP

mperham:
I get lots of reports of "stuck" Sidekiq jobs due to email delivery. SMTP servers in the wild are notoriously unreliable, e.g. spam honeypots can leave TCP connections lingering. We need network connection timeouts by default.


mperham:
See also rails/rails#41248

bf4:
@jeremy Looks fine to me

PR #1428, opened 2021-02-25
Use renamed main branch for rails in Gemfile

nevans:


c960657:
Why not just reference to the official released gem with just `gem 'activesupport'`? Add dependencies on other project's development branches is very unusual.

PR #1435, opened 2021-03-17
SMTP: allow disabling starttls_auto since it's now true by default in Ruby 3

### LINKED ISSUE #1434: Setting enable_starttls_auto to false will not work with Ruby 3
In [SMTP](https://github.com/mikel/mail/blob/master/lib/mail/network/delivery_methods/smtp.rb) class, there is a setting `:enable_starttls_auto`  which can be true or false. By default this setting is `true`. 

In Ruby 2.x, starttls is false, in Ruby 3 :auto.

The problem is that in[ delivery_methods smtp file](https://github.com/mikel/mail/blob/master/lib/mail/network/delivery_methods/smtp.rb ) , in method `build_smtp_session` there is no way to disable starttls, rather there is just a check to enable it, so starttls is always set to auto or true.

jeremy:
Passing enable_tls/starttls/starttls_auto: false now explicitly disables
these options. They used to be all false by default, so they could only
be turned ON. So we ignored turning them OFF!

We now disable_tls/starttls/starttls_auto when they're set to false.
Leaving them set to nil inherits the upstream default.

Fixes #1434.

### DIFF
--- a/lib/mail/network/delivery_methods/smtp.rb
+++ b/lib/mail/network/delivery_methods/smtp.rb
@@ -111,17 +111,33 @@ module Mail
 
       def build_smtp_session
         Net::SMTP.new(settings[:address], settings[:port]).tap do |smtp|
-          if settings[:tls] || settings[:ssl]
-            if smtp.respond_to?(:enable_tls)
+          tls = settings[:tls] || settings[:ssl]
+          if !tls.nil?
+            case tls
+            when true
               smtp.enable_tls(ssl_context)
+            when false
+              smtp.disable_tls
+            else
+              raise ArgumentError, "Unrecognized :tls value #{settings[:tls].inspect}; expected true, false, or nil"
             end
-          elsif settings[:enable_starttls]
-            if smtp.respond_to?(:enable_starttls)
+          elsif settings.include?(:enable_starttls) && !settings[:enable_starttls].nil?
+            case settings[:enable_starttls]
+            when true
               smtp.enable_starttls(ssl_context)
+            when false
+              smtp.disable_starttls
+            else
+              raise ArgumentError, "Unrecognized :enable_starttls value #{settings[:enable_starttls].inspect}; expected true, false, or nil"
             end
-          elsif settings[:enable_starttls_auto]
-            if smtp.respond_to?(:enable_starttls_auto)
+          elsif settings.include?(:enable_starttls_auto) && !settings[:enable_starttls_auto].nil?
+            case settings[:enable_starttls_auto]
+            when true
               smtp.enable_starttls_auto(ssl_context)
+            when false
+              smtp.disable_starttls_auto
+            else
+              raise ArgumentError, "Unrecognized :enable_starttls_auto value #{settings[:enable_starttls_auto].inspect}; expected true, false, or nil"
             end
           end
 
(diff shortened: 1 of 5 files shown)


HParker:
@jeremy this change worked for us at GitHub with net-smtp version `0.2.1`.

Thanks for finding this!

lib/mail/network/delivery_methods/smtp.rb
```diff
@@ -111,17 +111,33 @@ def start_smtp_session(&block)
...
+              raise ArgumentError, "Unrecognized :enable_starttls value #{settings[:enable_starttls].inspect}; expected true, false, or nil"
             end
-          elsif settings[:enable_starttls_auto]
-            if smtp.respond_to?(:enable_starttls_auto)
+          elsif settings.include?(:enable_starttls_auto) && !settings[:enable_starttls_auto].nil?
+            case settings[:enable_starttls_auto]
+            when true
               smtp.enable_starttls_auto(ssl_context)
+            when false
+              smtp.disable_starttls_auto
```
HParker, line 138: I think this should be `disable_starttls`. I can't find a reference to `disable_starttls_auto` in smtp, but I might be missing it.

ahorek:
@jeremy any update? this doesn't affect just Ruby 3, but even older Rubies with a recent net-smtp gem

without this patch, there's no way to disable ```starttls_auto```, even with explicit settings:
```
default:
  email_delivery:
    delivery_method: :smtp
    smtp_settings:
      tls: false
      enable_starttls_auto: false
      ....
```

seb-sykio:
+1

ahorek:
@mikel just a reminder, could we include this bug-fix into the 2.8 release? I can rebase if necessary, thanks!

mkllnk:
@hoeni's workaround helped us two. This is how it looks like in production code:
```patch
diff --git a/config/environments/production.rb b/config/environments/production.rb
index c1c2f442b..f215f49ce 100644
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -43,10 +43,10 @@ Fairfood::Application.configure do
 
   config.action_mailer.raise_delivery_errors = true
 
-  # Disable unnecessary TLS for local connection
+  # Disable TLS verification for local connection
   config.action_mailer.smtp_settings = {
     address: 'localhost',
-    enable_starttls_auto: false,
+    openssl_verify_mode: OpenSSL::SSL::VERIFY_NONE,
   }
 
   # Enable threaded mode
```

It avoids `hostname "localhost" does not match the server certificate (OpenSSL::SSL::SSLError)`.

lib/mail/network/delivery_methods/smtp.rb
GChuf: indeed, this **should** be disable_starttls - patch didn't work until I changed this line.

GChuf:
I forked 2-8-stable, added jeremy's patch and fixed "disable_starttls" flag.
You can clone it from here: https://github.com/GChuf/mail/

After that, build and install the gem locally:
_cd mail
gem build mail.gemspec
gem install mail -v2.8.0.pat_ch

omontigny:
Hello
Could someone please help me to troubleshhoot/configure sending email with starttls please ?
* Ruby : 3.1.2
* Rails : 6.1.7
* net-smtp: 0.3.3
* docker image : 3.1-slim
* I put my pem certificates files in /etc/ssl/certs folder 

without starttls it's  OK
config.action_mailer.smtp_settings = {
      address:              ENV.fetch('MAIL_SMTP_SERVER', 'smtp.svcs.dc1.xgs-france.com'),
      port:                 ENV.fetch('MAIL_SMTP_PORT', '25'),
      domain:               ENV.fetch('MAIL_DOMAIN','xgs-france.com'),
 
      openssl_verify_mode: OpenSSL::SSL::VERIFY_NONE
    }

mikel:
Hey there @ahorek please go ahead and do a new PR with this working against master and I'll release 2.9 shortly.

eval:
Fixed by https://github.com/mikel/mail/pull/1536

PR #1437, opened 2021-04-16
Fix multipart parsing

felipec:
When there's a LF format, and non-ascii characters the parsing doesn't
work.

### DIFF
--- a/lib/mail/body.rb
+++ b/lib/mail/body.rb
@@ -272,7 +272,7 @@ module Mail
       parts_regex = /
         (?:                    # non-capturing group
           \A                |  # start of string OR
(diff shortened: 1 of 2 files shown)


felipec:
This fixes #1438.

toofar:
Older PR here https://github.com/mikel/mail/pull/1437

felipec:
@mikel Hello? This is an obviously correct fix.

mikel:
Thanks @felipec and @toofar and @sebbASF I've created a new PR against Master here https://github.com/mikel/mail/pull/1511 using the original PR that has the test coverage

PR #1444, opened 2021-08-21
Support non-ASCII addresses

c960657:
The first internationalized domains (IDNs) were registered around 2003. It took more than a decade for web browsers to gain sufficient support to make these domains generally useful, but today you can announce an IDN domain to the general public and expect them to be able to connect to your website.

This is not the case for email. The standards for using IDNs for email came much later and they are still not supported by many email clients and software libraries. Let's fix that, so that eventually – many years from now – people can use their non-ASCII address everywhere like any other email address.

[RFC 6531](https://datatracker.ietf.org/doc/html/rfc) describes how to handle non-ASCII email addresses on the SMTP level using the `SMTPUTF8` SMTP extension. Unfortunately, `Net::SMTP` does not support passing parameters for `RCPT TO` which is required for `SMTPUTF8`. I have submitted [a patch](https://bugs.ruby-lang.org/issues/17047) for this, but the maintainers do not seem interested.

But even without `SMTPUTF8` support we can still get far:

With this PR, the domain of the SMTP envelope sender and recipient is IDNA-encoded, making the email address appear like any other ASCII email to the SMTP server. The SMTP envelope sender and recipient are never shown to end-users, so there is no UX reason not to do this encoding. Even if the SMTP server has support for non-ASCII domains, there is no drawback from doing the IDNA-encoding in the client – the two encodings are equivalent from a routing perspective.

This PR does not affect email addresses with non-ASCII characters in the local part. But even today, they might just work – not because email servers intentionally support internationalized email addresses, but simply because they pass on the local part verbatimly to the remote server.

If my PR for `Net::SMTP` is eventually accepted, this library could skip the IDNA-encoding and instead pass the `SMTPUTF8` parameter for servers announcing support for this extension. At least for addresses with a non-ASCII local part, this would probably improve deliverability.

### DIFF
--- a/lib/mail/elements/address.rb
+++ b/lib/mail/elements/address.rb
@@ -71,6 +71,18 @@ module Mail
       end
     end
 
+    # Returns the address that is in the address itself with domain IDNA-encoded.
+    #
+    #  a = Address.new('Mikel Lindsaar (My email address) <mikel@tést.lindsaar.net>')
+    #  a.address_idna #=> 'mikel@xn--tst-bma.lindsaar.net'
+    def address_idna
+      if d = domain_idna
+        "#{local}@#{d}"
+      else
+        local
+      end
+    end
+
     # Provides a way to assign an address to an already made Mail::Address object.
     #
     #  a = Address.new
@@ -118,6 +130,17 @@ module Mail
     def domain(output_type = :decode)
       parse unless @parsed
       Encodings.decode_encode(strip_all_comments(get_domain), output_type) if get_domain
+    end
+
+    # Returns the domain part (the right hand side of the @ sign in the email address) of
+    # the address in IDNA encoding.
+    #
+    #  a = Address.new('Mikel Lindsaar (My email address) <mikel@tést.lindsaar.net>')
+    #  a.domain_idna #=> 'xn--tst-bma.lindsaar.net'
+    def domain_idna
+      if d = domain
+        Encodings.idna_encode(d)
+      end
     end
 
     # Returns an array of comments that are in the email, or nil if there
--- a/lib/mail/smtp_envelope.rb
+++ b/lib/mail/smtp_envelope.rb
@@ -43,12 +43,16 @@ module Mail
 
     private
       def validate_addr(addr_name, addr)
-        if addr.bytesize > MAX_ADDRESS_BYTESIZE
-          raise ArgumentError, "SMTP #{addr_name} address may not exceed #{MAX_ADDRESS_BYTESIZE} bytes: #{addr.inspect}"
-        end
-
         if /[\r\n]/ =~ addr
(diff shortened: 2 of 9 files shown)


lib/mail/elements/address.rb
```diff
@@ -71,6 +71,18 @@ def address(output_type = :decode)
...
+    #
+    #  a = Address.new('Mikel Lindsaar (My email address) <mikel@tést.lindsaar.net>')
+    #  a.address_idna #=> 'mikel@xn--tst-bma.lindsaar.net'
+    def address_idna
+      if d = domain_idna
+        "#{local}@#{d}"
+      else
+        local
+      end
+    end
```
mikel, line 84: Doesn't this result in two separate cases?  You'll either return the local + domain or just the local.  Wouldn't it be better to just return the local here and put the domain_idna check in the same area where we return the full address (ie, local plus domain) ?

lib/mail/smtp_envelope.rb
```diff
@@ -43,14 +43,18 @@ def message=(message)
...
-          raise ArgumentError, "SMTP #{addr_name} address may not exceed #{MAX_ADDRESS_BYTESIZE} bytes: #{addr.inspect}"
-        end
-
         if /[\r\n]/ =~ addr
           raise ArgumentError, "SMTP #{addr_name} address may not contain CR or LF line breaks: #{addr.inspect}"
         end
 
+        if !addr.ascii_only? && Encodings.idna_supported?
+          addr = Address.new(addr).address_idna
+        end
```
mikel, line 52: Again, could the `Address` class handle this without having to spread the idna checks in the code?

mikel (changes requested):
Overall, excellent work @c960657 Thank you! I know it's been a long time since you wrote it, but I've left some comments around the PR, I think we could tighten this up some more, let me know if you think otherwise.

lib/mail/elements/address.rb
c960657: I'm not sure I follow. This method just follows the same pattern as `Address#address`, i.e. it does not add `@` if the address does not include a domain.

c960657:
After this PR was filed, an update to net-smtp (see https://github.com/ruby/net-smtp/pull/34) has allowed us to use the  `SMTPUTF8` extension. I have added support for this. If the SMTP server supports this extension, the simpleidn gem is not needed.

c960657:
Hmm, I'm not sure why the tests are failing. It seems related to Psych (v5.0.0 was released 2 days ago).

lib/mail/elements/address.rb
c960657: @mikel Would you elaborate? :-)

c960657: @mikel Ping :-)

PR #1446, opened 2021-09-15
fix smtp timeouts

ahorek:
1/ fixes a test failure introduced by https://github.com/mikel/mail/pull/1427

2/ some old rubies don't support open_timeout


ahorek:
closed in favor of https://github.com/mikel/mail/pull/1440

PR #1452, opened 2021-10-11
Handle question mark in encoded text

lewispb:
This PR updates the ENCODED_VALUE regex to handle an iso-8859-1 encoded value with a question mark in the middle of the text. This is an example string:

`=?iso-8859-1?Q?Hello World?_-_How are you??=`

I've also refactored the regexes to use the free-spacing option. This lets us comment each line of the regex improving readability.

### DIFF
--- a/lib/mail/constants.rb
+++ b/lib/mail/constants.rb
@@ -46,10 +46,10 @@ module Mail
       \?\=     # literal ?=
     }mix # m is multi-line, i is case-insensitive, x is free-spacing
 
-    FULL_ENCODED_VALUE = %r{
+    FULL_ENCODED_VALUE = %r{ # Identical to ENCODED_VALUE but captures the whole rather than components of
       (
-        \=\?
-        [^?]+
+        \=\?    # literal =?
+        [^?]+   #
         \?      # literal ?
         [QB]    # either a "Q" or a "B"
         \?      # literal ?


lib/mail/constants.rb
```diff
@@ -35,8 +35,27 @@ module Constants
...
     TOKEN_UNSAFE  = /[#{Regexp.quote tspecial}#{control}#{sp}]/n
-    ENCODED_VALUE = /\=\?([^?]+)\?([QB])\?[^?]*?\?+\=/mi
-    FULL_ENCODED_VALUE = /(\=\?[^?]+\?[QB]\?[^?]*?\?+\=)/mi
+
+    ENCODED_VALUE = %r{
+      \=\?     # literal =?
+      ([^?]+)  #
+      \?([QB]) # either a "Q" or a "B"
+      \?       # literal ?
+      [^?]?.*? #
```
lewispb, line 44: This is only part that's changed. Before `[^?]*?` after `[^?]?.*?`.

I must admit I was "inspired" by the python implementation here: https://github.com/python/cpython/blob/main/Lib/email/header.py#L41

lib/mail/constants.rb
```diff
@@ -35,8 +35,27 @@ module Constants
...
+    }mix # m is multi-line, i is case-insensitive, x is free-spacing
+
+    FULL_ENCODED_VALUE = %r{
+      (
+        \=\?
+        [^?]+
+        \?
+        [QB]
+        \?
+        [^?]*?
```
jeremy, line 55: Should this be changed to match as well?

lib/mail/constants.rb
jeremy: Can we ditch the `[^?]?` portion? The `.*?` would capture that as well.

lewispb: Ah, yes! Good point. Done.

lib/mail/constants.rb
```diff
@@ -35,8 +35,28 @@ module Constants
...
+      \=\?     # literal =?
+      ([^?]+)  #
+      \?       # literal ?
+      ([QB])   # either a "Q" or a "B"
+      \?       # literal ?
+      .*?      # lazily match all characters
+      \?\=     # literal ?=
+    }mix # m is multi-line, i is case-insensitive, x is free-spacing
+
+    FULL_ENCODED_VALUE = %r{ # Identical to ENCODED_VALUE but captures the whole rather than components of
```
jeremy, line 49: 🙏🏼

PR #1469, opened 2021-12-20
Handle too long dates passed to DateTime.parse

rosa:
Since [Ruby 3.0.3](https://www.ruby-lang.org/en/news/2021/11/24/ruby-3-0-3-released/), this method raises an `ArgumentError` with a different message from the invalid date message, as mitigation for [CVE-2021-41817](https://www.ruby-lang.org/en/news/2021/11/15/date-parsing-method-regexp-dos-cve-2021-41817/).

```
ArgumentError: string length (150) exceeds the limit 128
```

This PR handles this error to return `nil`, just like when the date is invalid.


lib/mail/fields/common_date_field.rb
```diff
@@ -4,26 +4,32 @@
...
       end
 
-      if datetime
-        datetime.strftime('%a, %d %b %Y %H:%M:%S %z')
-      else
-        string
+      def with_invalid_date_time_error_handling(&block)
+        yield
+      rescue ArgumentError => e
+        raise unless e.message =~ /\A(invalid date|string length \(\d+\) exceeds the limit \d+)\z/
```
rosa, line 32: Perhaps this should live in its own method (a method to parse dates that return `nil` when invalid or too long)  in `Mail::Utilities` so it's used from here and from `ReceivedElement`.

mikel: Yes, please move it into utilities.

lib/mail/elements/received_element.rb
```diff
@@ -27,7 +27,7 @@ def to_s(*args)
       def datetime_for(received)
         ::DateTime.parse("#{received.date} #{received.time}")
       rescue ArgumentError => e
-        raise e unless e.message == 'invalid date'
+        raise unless e.message =~ /\A(invalid date|string length \(\d+\) exceeds the limit \d+)\z/
```
mikel, line 30: Let's call the utility method you will create.  One place to keep this up to date makes sense.

mikel (changes requested):
Overall nice catch, thanks, please move to utility as you suggested

rosa:
Thanks a lot for the review, @mikel! I'll get these changes implemented this week, I hope.

rosa:
Implemented the changes requested 😊

zealot128:
Unfortunately, that is still an issue in 2024 with Mail 2.8.1 on 3.3.0.

That exception also breaks the POP functionality, as it crashes without option to rescue the individual mail:

```
mails = Mail.all delete_after_find: true

ArgumentError: string length (226) exceeds the limit 128 (ArgumentError)

          datetime = ::DateTime.parse(stripped)
                                      ^^^^^^^^
  from mail (2.8.1) lib/mail/fields/common_date_field.rb:17:in `parse'
  from mail (2.8.1) lib/mail/fields/common_date_field.rb:17:in `normalize_datetime'
  from mail (2.8.1) lib/mail/fields/common_date_field.rb:31:in `initialize'
  from mail (2.8.1) lib/mail/fields/common_field.rb:12:in `new'
  from mail (2.8.1) lib/mail/fields/common_field.rb:12:in `parse'
```

Any chance, this PR gets merged? 
For now, we need manual intervention and delete the mail from the inbox.

pdg137:
This is still an issue on the current master branch d1d65b370b109b98e673a934e8b70a0c1f58cc59. A long date field will crash the mail gem.  I'm working around it with a monkeypatch like this:

```ruby
require 'mail/fields/common_date_field' # must require this before overriding the method below

module Mail
  class CommonDateField < NamedStructuredField
    def initialize(value = nil, charset = nil)
      value = value && value[..127] # fix bug with long date fields, see https://github.com/mikel/mail/pull/1469
      super self.class.normalize_datetime(value), charset
    end
  end
end
```

PR #1470, opened 2021-12-24
Adding a part should not reset the mail's charset to nil

railsbob:
The PR https://github.com/mikel/mail/pull/1145 introduced a [breaking change](https://github.com/mikel/mail/pull/1145/files#diff-ea29154b6b75ca19a09c77672897b2673bc6f691d9f6e6ed81b057c1acb96023L2088) in `master`, where adding parts to a `Mail` resets the default `charset` to nil. 

This PR restores the behaviour from [`2-7-stable`](https://github.com/mikel/mail/blob/7c43c84c16f017e0ff5e5c9962f6a1d842301ee3/lib/mail/message.rb#L2087) and allows us to use the edge version of `mail` with rails 7 without any patch.

cc @jeremy

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -1496,7 +1496,7 @@ module Mail
     # Returns the character set defined in the content type field
     def charset
       if @header
-        has_content_type? ? content_type_parameters['charset'] : @charset
+        has_content_type? && !multipart? ? content_type_parameters['charset'] : @charset
       else
         @charset
       end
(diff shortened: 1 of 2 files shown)


lib/mail/message.rb
```diff
@@ -1496,7 +1496,7 @@ def mime_type
     # Returns the character set defined in the content type field
     def charset
       if @header
-        has_content_type? ? content_type_parameters['charset'] : @charset
```
railsbob, line 1499: Note: `multipart` `content-type` [does not have a `charset` parameter](https://github.com/mikel/mail/pull/1145), so effectively `content_type_parameters['charset']` was returning a `nil`.

mikel:
Thanks @railsbob I've merged this into master.  Any chance you could do another PR request against 2-8-stable?

mikel:
Never mind, did it here: https://github.com/mikel/mail/pull/1507  Thanks!

railsbob:
Thanks @mikel 🙌

PR #1472, opened 2022-02-16
Add Ruby 3.1 support

deivid-rodriguez:


lib/mail/encodings.rb
```diff
@@ -65,8 +65,7 @@ def Encodings.transcode_charset(str, from_charset, to_charset = 'UTF-8')
     #    param_encode_language 'jp'
     #  end
     #
-    # The character set used for encoding will either be the value of $KCODE for
-    # Ruby < 1.9 or the encoding on the string passed in.
+    # The character set used for encoding will either be the encoding on the string passed in.
```
betesh-at-data-axle, line 68:
```suggestion
    # The character set used for encoding will be the encoding on the string passed in.
```

eregon:
I was wondering why the truffleruby job gets cancelled, and found https://github.com/deivid-rodriguez/mail/actions/runs/1853944108
```
ruby (flavorjones/truffleruby:stable)
The job running on runner Hosted Agent has exceeded the maximum execution time of 5 minutes.
ruby (flavorjones/truffleruby:stable)
The operation was canceled.
```
So it seems because of the rather low `timeout-minutes`.
There are also some upcoming truffleruby changes to speed it up in CI and use less memory.
Anyway, seems OK to remove from now, that Docker image was also picking a older TruffleRuby release it seems.

And potentially the CI could (later) run outside Docker now that it doesn't require such ancient Ruby versions like 1.8, which would make the CI faster as well.

.github/workflows/test.yml
```diff
@@ -2,43 +2,41 @@ name: RSpec tests
...
+           timeout: 5
+         - ruby: truffleruby
+           timeout: 60
     steps:
     - uses: actions/checkout@v2
-    - name: Install bundler ${{ matrix.bundler-version }}
-      run: gem install bundler ${{ matrix.bundler-version && format('{0} {1}', '-v', matrix.bundler-version) || '' }}
-    - name: Install dependencies
-      run: bundle install
+    - uses: ruby/setup-ruby@v1.96.0
```
eregon, line 33: Let's use `@v1` and remove truffleruby for now, I guess this was tried to address https://github.com/deivid-rodriguez/mail/runs/5230121239?check_suite_focus=true?

That failure is an unfortunate consequence of depending on default gems explicitly, and the existing TruffleRuby release can do nothing about this (truffleruby-head doesn't have the issue, it just always uses the stdlib version). Well, it'd be possible to modify the upstream gem but it's unclear what's best at this point (ignore the gem & use the stdlib, or use the gem C ext but that's slower for some cases, or integrate in some other way).

deivid-rodriguez: There's something definitely wrong in `ruby/setup-ruby`. Bundle install is working perfect with bundler 2.3.7 on truffleruby. So I tried to use it by specifying `bundler: 2.3.7`, and `ruby/setup-ruby` ignored my explicit choice and still used an old Bundler. Version 1.96.0 does not have that bug.

eregon: It's due to https://github.com/ruby/setup-ruby/releases/tag/v1.97.0 / https://github.com/oracle/truffleruby/issues/2586 / https://github.com/rake-compiler/rake-compiler/issues/202.
`rake-compiler` started to mutate RbConfig in a way RubyGems can't find where to insert gems anymore.
One fix is to `@gem_prelude_index` but it can't be done for an existing release.
Using the Bundler shipped with TruffleRuby avoids multiple Bundler version and the problem, so that's the workaround for 22.0.
The output does say `Using Bundler 2 shipped with truffleruby-22.0.0.2` but doesn't explain why, should probably link to the issue.

deivid-rodriguez: Alright, but if I explicitly say the version of Bundler that I want because I'm having issues with the default version, and not hitting the issues it's preventing (because of not using `rake-compiler`, for example), shouldn't that be respected?

eregon: Unsure, because e.g. Bundler 1 is enforced on old Rubies (no mater the bundler input) because Bundler 2 wouldn't work or would just lead to confusing errors. Similar here, although it is less clear because latest Bundler could potentially work, if no `rake-compiler` involved, but that's not easy to find out (before the `bundle lock`).
I hate the situation, but it's also crazy rake-compiler (and RubyGems too, the code is copied from RubyGems) mutate RbConfig and there seem to be limited workarounds for that besides not mutating anymore.

deivid-rodriguez:
This is now green and I'm happy with it!

.github/workflows/test.yml
deivid-rodriguez: I really appreciate trying to provide a good default experience by adding workarounds for external issues, but the current behavior feels way too much to me. In any case, I'm happy sticking to 1.96 here until the next release of truffleruby!

simi (approved):

wolak88:
Is there anything else TBD before merging that to the gem itself?

Thanks for the great job!

simi:
@mikel @jeremy any chance to get some time to check this? Is there anything I can help with to help maintaining this gem?

eregon:
@deivid-rodriguez FWIW, I think there is no need to depend on the io-wait gem, is there?
Also for context: https://bugs.ruby-lang.org/issues/18591#note-7

Gemfile
```diff
@@ -34,3 +34,5 @@ elsif RUBY_VERSION >= '1.9'
 else
   gem 'ruby-debug', :platforms => :mri
 end
+
+gem "psych", "< 4"
```
voxik, line 17: I wish this got proper fix.

deivid-rodriguez: Be my guest.

voxik: Probably something like:

~~~
$ git diff
diff --git a/lib/mail/message.rb b/lib/mail/message.rb
index f5176ac8..95dfb7fe 100644
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -1841,7 +1841,7 @@ module Mail
     end
 
     def self.from_yaml(str)
-      hash = YAML.load(str)
+      hash = YAML.load(str, :permitted_classes => [Mail::Body, Mail::SMTP, Symbol])
       m = self.new(:headers => hash['headers'])
       hash.delete('headers')
       hash.each do |k,v|
diff --git a/spec/mail/message_spec.rb b/spec/mail/message_spec.rb
index 033edf2b..0bdfbef6 100644
--- a/spec/mail/message_spec.rb
+++ b/spec/mail/message_spec.rb
@@ -196,7 +196,7 @@ describe Mail::Message do
 
       it "should serialize the basic information to YAML" do
         yaml = @yaml_mail.to_yaml
-        yaml_output = YAML.load(yaml)
+        yaml_output = YAML.respond_to?(:unsafe_load) ? YAML.unsafe_load(yaml) : YAML.load(yaml)
         expect(yaml_output['headers']['To']).to       eq "someone@somewhere.com"
         expect(yaml_output['headers']['Cc']).to       eq "someoneelse@somewhere.com"
         expect(yaml_output['headers']['Subject']).to  eq "subject"
@@ -214,7 +214,7 @@ describe Mail::Message do
       it "should serialize a Message with a custom delivery_handler" do
         @yaml_mail.delivery_handler = DeliveryAgent
         yaml = @yaml_mail.to_yaml
-        yaml_output = YAML.load(yaml)
+        yaml_output = YAML.respond_to?(:unsafe_load) ? YAML.unsafe_load(yaml) : YAML.load(yaml)
         expect(yaml_output['delivery_handler']).to eq "DeliveryAgent"
       end
 
@@ -226,7 +226,7 @@ describe Mail::Message do
 
       it "should not deserialize a delivery_handler that does not exist" do
         yaml = @yaml_mail.to_yaml
-        yaml_hash = YAML.load(yaml)
+        yaml_hash = YAML.respond_to?(:unsafe_load) ? YAML.unsafe_load(yaml) : YAML.load(yaml)
         yaml_hash['delivery_handler'] = "NotARealClass"
         deserialized = Mail::Message.from_yaml(yaml_hash.to_yaml)
         expect(deserialized.delivery_handler).to be_nil
@@ -234,7 +234,7 @@ describe Mail::Message do
 
       it "should deserialize parts as an instance of Mail::PartsList" do
         yaml = @yaml_mail.to_yaml
-        yaml_hash = YAML.load(yaml)
+        yaml_hash = YAML.respond_to?(:unsafe_load) ? YAML.unsafe_load(yaml) : YAML.load(yaml)
         deserialized = Mail::Message.from_yaml(yaml_hash.to_yaml)
         expect(deserialized.parts).to be_kind_of(Mail::PartsList)
       end
~~~

While this works running just the message_spec.rb,  I'm observing the following error when running the complete test suite:

~~~
  1) Mail::Message initialization YAML serialization should handle multipart mail
     Failure/Error: hash = YAML.load(str, :permitted_classes => [Mail::Body, Mail::SMTP, Symbol])
     
     Psych::DisallowedClass:
       Tried to load unspecified class: MyTestDeliveryMethod
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/class_loader.rb:99:in `find'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/class_loader.rb:28:in `load'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/to_ruby.rb:424:in `resolve_class'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/to_ruby.rb:213:in `visit_Psych_Nodes_Mapping'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/visitor.rb:30:in `visit'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/visitor.rb:6:in `accept'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/to_ruby.rb:35:in `accept'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/to_ruby.rb:345:in `block in revive_hash'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/to_ruby.rb:343:in `each'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/to_ruby.rb:343:in `each_slice'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/to_ruby.rb:343:in `revive_hash'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/to_ruby.rb:167:in `visit_Psych_Nodes_Mapping'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/visitor.rb:30:in `visit'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/visitor.rb:6:in `accept'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/to_ruby.rb:35:in `accept'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/to_ruby.rb:318:in `visit_Psych_Nodes_Document'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/visitor.rb:30:in `visit'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/visitor.rb:6:in `accept'
     # /usr/share/gems/gems/psych-4.0.3/lib/psych/visitors/to_ruby.rb:35:in `accept'
     # ./lib/mail/message.rb:1870:in `from_yaml'
     # ./lib/mail/message.rb:1883:in `block (2 levels) in from_yaml'
     # ./lib/mail/message.rb:1883:in `map'
     # ./lib/mail/message.rb:1883:in `block in from_yaml'
     # ./lib/mail/message.rb:1873:in `each'
     # ./lib/mail/message.rb:1873:in `from_yaml'
     # ./spec/mail/message_spec.rb:240:in `block (4 levels) in <top (required)>'
~~~

Undefining the `MyTestDeliveryMethod` somewhere would be probably the right thing to do.

simi: I'm on it.

deivid-rodriguez: Ideally it should be part of a PR that gets the base CI green (that was the point of restricting the version in this branch). But I think it's fine to do it here, since two unattended PRs seems enough already 😅, so feel free to send a PR to my fork 👍.

deivid-rodriguez: Oh, I see that you already created #1476, then it's fine, if that one gets attended & merged first, I'll rebase this one and remove the upper bound from the Gemfile.

mikel:
Let's get https://github.com/mikel/mail/pull/1476 passing CI and merged first, then we can rebase this and move forward @deivid-rodriguez @simi and @voxik

simi:
:pray: time to rebase @deivid-rodriguez

deivid-rodriguez:
This should be ready now!

Gemfile
```diff
@@ -13,3 +13,6 @@ gem 'jruby-openssl', :platforms => :jruby
 gem 'mini_mime'
 
 gem 'byebug', :platforms => :mri
+
+gem "io-wait", ">= 0.2.2.pre1"
+gem "strscan", ">= 3.0.2.pre1"
```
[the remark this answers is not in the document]
eregon: Ah JRuby needs it, and `strscan` is pulled in as a gem from one of the net-* gems I guess:
https://github.com/mikel/mail/runs/6114147775?check_suite_focus=true

`io-wait` shouldn't be a dependency of any of the `net-*` gems anymore, but unsure if there were releases with that.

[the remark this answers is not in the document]
deivid-rodriguez: Yeah, JRuby needs the prereleases but Bundler by default prefers final releases which don't really work. So we need to force them via Gemfile. It doesn't seem like a big deal, it can be removed once there are final releases of those.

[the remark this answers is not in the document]
eregon: `io/wait` is already not needed though, could you remove that?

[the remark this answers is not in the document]
deivid-rodriguez: Right, `io-wait` was necessary when I first create the PR, but it's no longer a transitive dependency. Let me remove it.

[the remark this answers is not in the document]
eregon: Thanks, for `strscan` it currently works by luck in TruffleRuby (see edited comment above), that'll be fixed for future strscan releases with https://github.com/ruby/strscan/pull/35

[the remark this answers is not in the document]
headius: We will not ship JRuby 9.4 with dependencies on the preview gems, so they should go away once we have worked out all issues related to the JRuby versions.

mikel:
Awesome work @deivid-rodriguez @voxik and everyone!

mikel:
@deivid-rodriguez  / @voxik / @eregon / @jeremy  https://rubygems.org/gems/mail/versions/2.8.0.rc1

jrochkind:
2.8.0rc1 was released April 22, a bit over 7 weeks.  It's working well for me. 

I have some projects that use code to automatically generate Rails apps for test builds. It would make it quite a lot easier for this system to generate Rails 6.1.0 apps under ruby 3.1 if a 2.8.0 final was released. 

@mikel appreciate if you can share any information on expected timeline or necessary pre-requisite steps to happen for a 2.8.0 final release?

johnnyshields:
@mikel Shall we do the final release of 2.8.0? This issue is causing a problem for many people as you can see in all the references above.

ghepting:
Is there a stable release with this out yet or just the rc?

ashishra0:
Folks, I recently upgraded to `ruby3.1.1` in my project which sends mail to users and hit a blocker, and to resolve that I need to add the `net/smtp` gem. Now my question is, do I still need to include that gem in my gemfile?

deivid-rodriguez:
Little has changed, unfortunately. Previously to the rc release, you needed to include `net/smtp` in your Gemfile. Now you need to either use that workaround or force the rc to be picked up by Bundler (with `gem "mail", ">= 2.8.0.rc1"` or similar).

Hopefully @mikel will release the final version soon.

johnnyshields:
@mikel any update on this?

PR #1475, opened 2022-03-12
Add `delivery_interceptors` method to `Mail` class to fetch registered interceptors

ghousemohamed:
Summary:

I've raised this PR in reference to this [PR](https://github.com/rails/rails/pull/44638) in rails and this [comment](https://github.com/rails/rails/pull/44638#issuecomment-1063326528) specifically in the PR.

This PR adds the `delivery_interceptors` method to the `Mail` class to fetch the registered interceptors. So instead of fetching like `Mail.class_variable_get(:@@delivery_interceptors)`, we can directly do `Mail.delivery_interceptors`.

### DIFF
--- a/lib/mail/mail.rb
+++ b/lib/mail/mail.rb
@@ -236,6 +236,11 @@ module Mail
     end
   end
 
+  # Returns a list of registered delivery interceptors.
+  def self.delivery_interceptors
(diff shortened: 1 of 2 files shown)


mikel:
Beautiful, thank you :)

PR #1476, opened 2022-03-23
Add Mail::YAML#load compatible with Psych 3.x and Psych 4.x.

simi:
This is 1:1 transition. It would be great to be able to use "safe" load, but since custom classes could be serialised, it is not easily possible for now.

Btw. the best would be to be able to migrate to Psych 4, but it seems impossible since EOL Rubies are supported.

@voxik @deivid-rodriguez

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -2,7 +2,7 @@
 # frozen_string_literal: true
 require 'mail/constants'
 require 'mail/utilities'
-require 'yaml'
+require 'mail/yaml'
 
 module Mail
   # The Message class provides a single point of access to all things to do with an
@@ -1841,7 +1841,7 @@ module Mail
     end
 
     def self.from_yaml(str)
-      hash = YAML.load(str)
+      hash = Mail::YAML.load(str)
       m = self.new(:headers => hash['headers'])
       hash.delete('headers')
       hash.each do |k,v|
--- a/lib/mail/yaml.rb
+++ b/lib/mail/yaml.rb
@@ -1,0 +1,16 @@
+require 'yaml'
+
+module Mail
+  module YAML
+    # unsafe loading compatible with Psych 3.x and Psych 4.x
+    if ::YAML.respond_to?(:unsafe_load)
+      def self.load(yaml)
+        ::YAML.unsafe_load(yaml)
+      end
+    else
+      def self.load(yaml)
+        ::YAML.load(yaml)
+      end
+    end
+  end
+end
--- a/spec/mail/message_spec.rb
+++ b/spec/mail/message_spec.rb
@@ -196,7 +196,7 @@ describe Mail::Message do
 
       it "should serialize the basic information to YAML" do
         yaml = @yaml_mail.to_yaml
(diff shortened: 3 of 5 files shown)


voxik:
Very nice. Thx. This fixes the compatibility. Since there is check for the [`delivery_handler`](https://github.com/mikel/mail/blob/6e5db59a349a2af9146d16d8f66cbd3b81cd905a/lib/mail/message.rb#L1849), I don't feel very bad about usa of the `unsafe_load`.

voxik:
> Since there is check for the [`delivery_handler`](https://github.com/mikel/mail/blob/6e5db59a349a2af9146d16d8f66cbd3b81cd905a/lib/mail/message.rb#L1849), I don't feel very bad about usa of the `unsafe_load`.

Hm, this is probably not very relevant :roll_eyes: Nevertheless, how about:

~~~
$ git diff
diff --git a/lib/mail/yaml.rb b/lib/mail/yaml.rb
index c0910855..b70da033 100644
--- a/lib/mail/yaml.rb
+++ b/lib/mail/yaml.rb
@@ -2,15 +2,23 @@ require 'yaml'
 
 module Mail
   module YAML
-    # unsafe loading compatible with Psych 3.x and Psych 4.x
-    if ::YAML.respond_to?(:unsafe_load)
-      def self.load(yaml)
-        ::YAML.unsafe_load(yaml)
-      end
-    else
-      def self.load(yaml)
-        ::YAML.load(yaml)
-      end
+    def self.load(yaml)
+      ::YAML.load(yaml, :permitted_classes => [
+        Symbol,
+
+        Mail::Body,
+
+        # Delivery methods as listed in mail/configuration.rb
+        Mail::SMTP,
+        Mail::Sendmail,
+        Mail::Exim,
+        Mail::FileDelivery,
+        Mail::SMTPConnection,
+        Mail::TestMailer,
+        Mail::LoggerDelivery,
+
+        Mail.delivery_method.class,
+      ])
     end
   end
 end
~~~

mikel:
Hey there @voxik thanks for the help on creating this. I just kicked off the workflows and looks like we have some failures on 1.9, 2.0, 2.2, 2.7 and latest.  Can you take a quick look at those and see if we can fix those up?  cc @deivid-rodriguez

mikel:
Sorry I'm late to the party :)  Happy to help get these 3.x PRs resolved and merged and a new gem released asap.

deivid-rodriguez:
My recall about current failures from working on #1472:

* The one on Ruby 2.7 is an existing flaky failure that happens randomly on different rubies. Restarting should "fix it".
* The failures on older rubies I think are already happening on master, and #1472 is dropping support anyways.

So I think this PR should be fine!

deivid-rodriguez:
@mikel I unified changes from #1472 and this PR to get a base green CI at https://github.com/mikel/mail/pull/1478. It would make this PR unnecessary (I kept proper attribution to @simi and @voxik at #1478).

Then I can rebase #1472 to take care of adding Ruby 3.1 support.

What do you think?

PR #1478, opened 2022-04-19
Require Ruby >= 2.5, add Psych 4 support, and get CI green

deivid-rodriguez:
I unified the changes from #1472 and #1476 intended to get a base green CI line.

I tweaked #1476 to use safe loading by default but still fully support Psych 3.

Link to green actions run on my fork: https://github.com/deivid-rodriguez/mail/runs/6076144108?check_suite_focus=true.


README.md
```diff
@@ -48,7 +47,7 @@ our documentation, add new features—up to you! Thank you for pitching in.
 
 ## Compatibility
 
-Mail supports Ruby 1.8.7+, including JRuby and Rubinius.
+Mail supports Ruby 2.5+, including JRuby and Rubinius.
```
simi, line 50: I think Rubinius is gone (at least at CI level). We can swap it with TruffleRuby.

```suggestion
Mail supports Ruby 2.5+, including JRuby and TruffleRuby.
```

lib/mail/yaml.rb
```diff
@@ -0,0 +1,30 @@
...
+        Mail::Exim,
+        Mail::FileDelivery,
+        Mail::SMTPConnection,
+        Mail::TestMailer,
+        Mail::LoggerDelivery,
+
+        Mail.delivery_method.class,
+      ]
+
+      if Gem::Version.new(Psych::VERSION) >= Gem::Version.new('3.1.0.pre1')
```
voxik, line 23: @mikel What is the mail support matrix? Is it worth of the version check if Ruby/Psych upstream was fine to abandon the backward compatibility?

mikel: Hey there @voxik - I don't have a standardized one at this point mainly because with the way I enabled older ruby support it wasn't a big deal to just keep it current and mail is mainly string handling.  At some point we gotta drop 1.8 and 1.9 though :)

voxik: Now I see that Ruby 2.5+ is mentioned on other places, which means Psych 3.0.2, so it is probably needed, unless it is reconsidered ...

deivid-rodriguez: We're defining that right now 😅.

My proposal is to keep support for Ruby 2.5 and higher, and since Ruby 2.5 ships with Psych 3.0.2, I added this for backwards compatibility. I would be 100% on board to also dropping Ruby 2.5 support, and in that case I believe it would be fine to also drop support for Psych 3.0.

Whatever @mikel feels is best seems ok to me!

mikel: I think we could match Ruby, for example https://www.ruby-lang.org/en/downloads/ states that 2.5.9 is in EOL.  So 2.5+ could be a reasonable base.

lib/mail.rb
```diff
@@ -9,26 +9,11 @@ module Mail # :doc:
...
-  end
-
-  if RUBY_VERSION >= "1.9.0"
-    require 'mail/version_specific/ruby_1_9'
-    RubyVer = Ruby19
-  else
-    require 'mail/version_specific/ruby_1_8'
-    RubyVer = Ruby18
-  end
+  require 'mail/version_specific/ruby_1_9'
```
mikel, line 12: @deivid-rodriguez is this needed any more if we are not supporting 1.9?

lib/mail/yaml.rb
deivid-rodriguez: Great, let's drop support for Ruby 2.5 and Psych 3.0 too, let me add that to this PR.

mikel: This will probably be a Mail v3 release, because it's "very" breaking to systems running older rubies.

lib/mail.rb
deivid-rodriguez: I think the module is still needed, but the `ruby_1_9` naming is now misleading since there's now a single implementation of these methods. I guess it could be renamed to something different, for example, `Mail::Utils`.

lib/mail/yaml.rb
voxik: BTW will dropping Ruby 2.5 support require major version bump?

lib/mail.rb
mikel: Please do implement.

README.md
```diff
@@ -48,7 +47,9 @@ our documentation, add new features—up to you! Thank you for pitching in.
 
 ## Compatibility
 
-Mail supports Ruby 1.8.7+, including JRuby and Rubinius.
+Mail supports Ruby 2.5+, including JRuby and TruffleRuby.
+
+As new versions of Ruby are released, Mail will be compatible with support for the "preview" and all "normal maintenance", "security maintenance" and the two most recent "end of life" versions listed at the [Ruby Maintenance Branches](https://www.ruby-lang.org/en/downloads/branches/) page.  Pull requests to assist in adding support for new preview releases are more than welcome.
```
mikel, line 52: @deivid-rodriguez @voxik I added this paragraph, seem reasonable to you?  @jeremy do you have any input?

lib/mail/yaml.rb
mikel: @voxik no I don't think so.  See my update of the README in this same PR for what I'm thinking.

README.md
deivid-rodriguez: Ah, alright, so we do keep Ruby 2.5 support, correct?

lib/mail.rb
deivid-rodriguez: There's a `Mail::Utilities` module already. I think it makes sense to move these methods there.

README.md
deivid-rodriguez: It seems reasonable to me, by the way!

voxik: That makes sense :+1: 

> Ah, alright, so we do keep Ruby 2.5 support, correct?

I read it the same way.

lib/mail.rb
deivid-rodriguez: Alright, for now I removed the `Ruby19` module and the `ruby_1_9.rb` file in favor of the `RubyVer` module (which existed already), and a `version_specific.rb` file. Still not great since most of the stuff at `version_specific.rb` is not really version specific anymore. But seems much better than before and it was an easy change.

lib/mail/yaml.rb
simi: I suggest to at least release last 2.x version with improvements keeping 2.5 support and later (in other PR) we can prepare 3.0 release removing additional bandaids and drop 2.5 support.

lib/mail.rb
deivid-rodriguez: I guess best would be to move everything to `Mail::Utilities`, I'll do it tomorrow first thing since now I have to leave the computer :)

lib/mail/yaml.rb
mikel: @simi I'll just release it as a the next 2.x version but drop support for rubies `< 2.5`

lib/mail.rb
deivid-rodriguez: Done!

mikel:
Awesome work @deivid-rodriguez - could @voxik or @simi please do one review through this and give it a thumbs up? This is a pretty major change and I would love another couple of eyes on it.  Then I'll merge and we'll release Mail 2.8.0.rc1 into a stable branch

deivid-rodriguez:
Thanks for having a look @voxik 👍. I'll rebase and get #1472 ready as soon as this is merged 💪.

simi (approved):

jeremy:
That maintenance/support policy sounds good. Does it square with the
existing policy of the extracted stdlib gems? Considering we'll be bound by
the support policies of those and future stdlib gem extractions.

deivid-rodriguez:
In my experience, the support policy of extracted stdlibs is not really established, but they are usually more aggressive than this when it comes to dropping support for old rubies. For example, net-imap 0.2.2 supports ruby 2.5+, while net-imap 0.2.3 (the latest version) supports ruby 2.6+.

mikel:
Looks good to me @deivid-rodriguez @voxik - let's get 1472 under control now.  Then I'll release 2.8.0.rc1

lib/mail.rb
```diff
@@ -9,26 +9,8 @@ module Mail # :doc:
...
-  end
-
-  if RUBY_VERSION >= "1.9.0"
-    require 'mail/version_specific/ruby_1_9'
-    RubyVer = Ruby19
-  else
-    require 'mail/version_specific/ruby_1_8'
-    RubyVer = Ruby18
-  end
-
```
jeremy, line 27: I'm afraid these are public constants that are broadly referenced, so removal will cause surprising downstream breakage: [RubyVer](https://github.com/search?q=%22Mail%3A%3ARubyVer%22&type=code), [Ruby18](https://github.com/search?q=%22Mail%3A%3ARuby18%22&type=code), [Ruby19](https://github.com/search?q=%22Mail%3A%3ARuby19%22&type=code)

A deprecation cycle would be sensible.

deivid-rodriguez: Also `Ruby18`, given that we're dropping Ruby 1.8 support? The link you provided does not show any meaningful results in the first pages for me.

PR #1481, opened 2022-05-02
Add configurable mime boundary generator

dugancathal:
Hello!

This PR aims to add the ability to customize the generator used for
mime-boundaries when creating mails. 

Recently, after converting our application to use ActionMailer (and thus, Mail),
my team found that several of our customers stopped receiving our emails due to
"malformed structures". Upon digging in, it appears their systems may be
flagging our emails because of the new email's mime-boundary containing an equal
sign (`=`). While the RFC states that this character is valid in the boundary
(provided it's quoted), we have found a couples instances noting that it's
merely safer to not include them at all.

[Example 1](https://www.oreilly.com/library/view/programming-internet-email/9780596802585/ch03s04.html)
[Example 2](https://docs.microsoft.com/en-us/exchange/troubleshoot/administration/multipart-mixed-mime-message-format)

Rather than request a change to the boundary generation algorithm here, we
instead make the algorithm configurable to hopefully prevent any impact to
current consumers.

Feedback is, of course, welcome on the code structures presented here and/or any
other alternative options that may help us diagnose the issue.

Thank you!


dugancathal:
Hey @mikel! Thanks for the review.

I'm not 100% sure I follow, so I'd like to clarify. Are you looking for me to:
1. Undo the change here and instead completely remove the `=` from the `generate_boundary` method? 
2. Update the `Mail.random_tag` method to, for instance, `gsub('=', '')` (or similar)?
3. Something else?

Looking at the code, I'm not sure that [2] would do anything as all the components of that string are numeric and will not include equal signs (`=`).

Just lemme know which you're lookin' for and I'll get this updated for release.

Cheers!

mikel:
Hey there @dugancathal - you mention the reason you created the custom mime boundry generator capability is because Mail is including a `=` in the mime boundry generated.

So what I'm proposing is:

1) Keep these changes (allowing people to override the mime boundry generator is a good thing, maybe they want a custom boundry for some reason)
2) ALSO modify the default mail mime boundry generator itself so that it does not include an `=` in it to remove the need for 99% of people to ever need to implement their own custom one (including you) :)

Make sense?

dugancathal:
Makes sense, @mikel. Made those changes. Lemme know if this needs any other work. :)

mikel:
OK thanks @dugancathal - any chance you could look into the spec failures?

dugancathal:
Hey folks, sorry for the radio silence on this. Is there anything y'all need from me? Or is it possible for me to rebase and re-run CI somehow to see if this is still an issue?

PR #1483, opened 2022-05-09
Remove unnecessary gemfile dependency on strscan

deivid-rodriguez:
The final version is out, so this is no longer necessary since Bundler will resolve to that automatically.


mikel:
Thanks @deivid-rodriguez - please add a changelog and I'll merge this.

deivid-rodriguez:
Hi @mikel!

The `Gemfile` file is not really used by end users of the mail gem, it's just useful for developing it. Do you still think a changelog note should be added?

mikel:
Fair call.

PR #1491, opened 2022-06-16
Lazy-load fields and elements

c960657:
Even though there are autoload rules specified for (nearly) all field classes, these are eager-loaded by `require 'mail'`. This defeats some of the purpose with using autoload.

I have used this little script to see how many files are included:

```
module Kernel
  alias_method :old_require, :require

  def require(file)
    loaded = old_require(file)
    puts file if loaded
    loaded
  end
end

$LOAD_PATH.unshift "#{__dir__}/lib"
require 'mail'
```

This PR removes the explicit loading files in `lib/mail/fields` and `lib/mail/elements`. This reduces the number files initially required from 95 to 42. The loading of these files is postponed until this gem is actually used to send a mail.

The performance impact of this is tiny, but if you use many gems, it all adds up.


lib/mail/field.rb
```diff
@@ -145,7 +109,12 @@ def split(raw_field) #:nodoc:
       end
 
       def field_class_for(name) #:nodoc:
-        FIELDS_MAP[name.to_s.downcase]
+        name = name.to_s.downcase
+        @fields_map ||= {}
+        @fields_map[name] ||= begin
+          klass = name.split('-').map(&:capitalize).join + "Field"
+          Mail.const_get(klass) rescue nil
+        end
```
mikel, line 117: Thanks @c960657 so much for this PR, I'd prefer to keep FEILDS_MAP though as , if we don't want to instantiate the class we could map the class names as a string, then in this method just call `Mail.const_get(FIELDS_MAP[name.to_s.downcase])`

Reason is for simplicity for future debugging and for class names that might not match the field header name for some reason in the future.

mikel: If you can modify the above and ideally not delete the FIELD_MAP_NAME hash as well, I'll merge this for the next release.

c960657: I just realised my latest commit wasn't quite right. I'll push an update later today.

c960657: Done. I hope this is what you intended.

The new code is a bit more explicit, but I have added tests to help keep the large arrays/hashes in sync.

mikel: Nice, thank you @c960657 :)

Issue #1493, opened 2022-06-27
Semver break in 2.8.0.rc1 - CheckDeliveryParams


johnnyshields:
Mail 2.8.0.rc1 removes the class `Mail::CheckDeliveryParams`. This is a public interface change which means that according to [Semver](https://semver.org/) either:

- Do a major version bump (3.0.0); or
- Re-add the class interface to ensure compatibility (can be marked `@deprecated`)

Otherwise, this may break some gems; I found that it broke the `mail-ses` gem which I maintain.
https://github.com/tablecheck/mail-ses/pull/4

mikel:
Thanks for finding this, will investigate and fix.

mikel:
Hi there @johnnyshields I think https://github.com/mikel/mail/pull/1502 should fix it for now.

PR #1502, opened 2022-11-30
Restoring Mail::CheckDeliverParam and deprecating

mikel:
Can't remove until our next major version release.

Resolves https://github.com/mikel/mail/issues/1493


mikel:
Reversed this as it should have gone into 2.8 not master

Issue #1497, opened 2022-08-05
The "Mime-Version" header capitalization is considered a spam signal by rspamd


borama:
I noticed that the mail gem [generates](https://github.com/mikel/mail/blob/641060598f8f4be14d79bad8d703e9f2967e1cdb/lib/mail/fields/mime_version_field.rb#L8) the MIME version header capitalized as `Mime-Version` as opposed to `MIME-Version` mentioned in the [RFC 2045](https://www.rfc-editor.org/rfc/rfc2045#section-4). 

While this spelling is technically [perfectly correct](https://stackoverflow.com/a/6143644/1544012), it is specifically matched by an [rspamd](https://www.rspamd.com) antispam [rule called `MV_CASE`](https://github.com/rspamd/rspamd/blob/17cb2c06fd5d7dea31c1f1215ee3d5ba3cabe5f3/rules/headers_checks.lua#L589) which effectively raises the spam score by 0.5 for all emails with such header. I traced down the [original commit](https://github.com/rspamd/rspamd/commit/5c669479a0e0630f822929714332b615f11210a6) in rspamd, it has been there for years and seems to be included in the default configuration, but other than that I found no info about why this rule was added.

I briefly checked a few mail service providers and found that the header capitalization varies a lot:
- GMail uses "MIME-Version"
- FastMail uses "Mime-Version"
- HEY uses "Mime-Version" (but, being a Rails app, perhaps it uses this gem to generate emails :) )

I'll be glad to open a PR if you're interested, thanks!

mikel:
Fixed in https://github.com/mikel/mail/pull/1503

borama:
Thanks @mikel for taking a look at this!

PR #1503, opened 2022-11-30
Update MIME-Version to have correct case per the RFC

mikel:
per https://github.com/mikel/mail/issues/1497

Should return `MIME-Version` not `Mime-Version` per [RFC 2045](https://www.rfc-editor.org/rfc/rfc2045#section-4)


dawagner:
Just wanted to say thanks for the fix, @mikel  :) I have some outgoing mails sent by zammad, using this gem, that are blocked, in part because of this. However, there hasn't been a new release since this was merged :(

dawagner:
Oops, sorry, I've just noticed that it *has* been release. I was misled by https://github.com/mikel/mail/blob/master/CHANGELOG.rdoc mentioning 2.9.0 as unreleased.

PR #1511, opened 2022-12-03

mikel:
Original PR https://github.com/mikel/mail/pull/1319

Conversion of body to CRLF does not help here because body is not ascii_only so does not get converted. Instead to search for CR?LF (optional CR) in which case conversion to CRLF is unnecessary.

### DIFF
--- a/.github/workflows/adhoc.yml
+++ b/.github/workflows/adhoc.yml
@@ -1,0 +1,80 @@
+name: Adhoc RSpec tests
+
+on:
+  workflow_dispatch:
+
+  pull_request:
+    branches: [ feature/parse_lf ]
+  push:
+    branches: [ feature/parse_lf ]
+
+jobs:
+  ruby:
+    name: ${{ matrix.ruby }} (timeout ${{ matrix.timeout }})
+    runs-on: ubuntu-latest
+    timeout-minutes: ${{ matrix.timeout }}
+    strategy:
+      fail-fast: false
+      matrix:
+       include:
+         - ruby: 2.5
+           timeout: 5
+         - ruby: 2.6
+           timeout: 5
+         - ruby: 2.7
+           timeout: 5
+         - ruby: '3.0'
+           timeout: 5
+         - ruby: 3.1
+           timeout: 5
+         - ruby: jruby
+           timeout: 10
+         - ruby: jruby-head
+           timeout: 10
+         - ruby: jruby-9.4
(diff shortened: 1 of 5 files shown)


mikel:
Thanks @felipec and @toofar and @sebbASF for this, the specs are failing in JRuby 9.2, if you'd like to help work out why, that would be grand and I'll get this merged.

mikel:
Hey there @sebbASF how annoying! Could you please go ahead and add the workflow_dispatch and I'm going to ping Charles and see if he has any ideas on this, because it's a core change to our parser, I want to try and get to the bottom of it.

headius:
In the current builds I can only see 9.2 errors. They range across several tests and I'm not sure if there's a common cause. If you can isolate them to specific reproductions the next step would be filing some JRuby issues.

If you can point me at the intermittent 9.3 tests I can have a look into that.

headius:
> However on JRuby, tests then fail because the method _native_parse cannot be found.

Aha! Ok so there's just some Psych 5 updates we need to make for JRuby.

headius:
The `_native_parse` error is due to a bad release of the Psych 5.0 gem: https://github.com/ruby/psych/issues/598

headius:
Psych 5.0.1 has been released, fixing the `_native_parse` issue.

https://rubygems.org/gems/psych/versions/5.0.1-java

mikel:
Awesome, thanks @sebbASF and @headius  :)

alexdean:
@mikel hi! i was looking for info about when i might see the change in the PR in an official `mail` release. looked for some info on the release process & didn't find anything.

context: i maintain the [as2 gem](https://github.com/alexdean/as2). AS2 messages are MIME messages, so we use `mail` to parse out the message body & digital signature. starting in mail 2.7.1, all line endings in message bodies are being converted to `\r\n`, even when the raw content had only `\n` line endings. This breaks signature verification.

```ruby
# before this PR...
require 'mail'
mail = Mail::Body.new("new\nbody")
mail.raw_source
# => "new\r\nbody"       <------- note additional "\r"

# after this PR...
require 'mail'
mail = Mail::Body.new("new\nbody")
mail.raw_source
# => "new\nbody" 
```

The change in this PR fixes the issue, so I'm hoping to find out when I can start using it in `as2`. If a release is some time away, can you suggest any way to patch this behavior in a relatively well-contained way?

PR #1512, opened 2022-12-03

mikel:
Forward port of https://github.com/mikel/mail/pull/1319

Conversion of body to CRLF does not help here because body is not ascii_only so does not get converted. Instead to search for CR?LF (optional CR) in which case conversion to CRLF is unnecessary.


mikel:
As the 2.8 branch doesn't have all the specs running automatically, waiting for https://github.com/mikel/mail/pull/1437 to be green before merging this.

mikel:
@sebbASF can you confirm there is nothing missing from https://github.com/mikel/mail/pull/1511 ?

PR #1514, opened 2022-12-03
use match?

ahorek:
if there's no backreference in use, we can optimize it https://github.com/fastruby/fast-ruby#stringmatch-vs-stringmatch-vs-stringstart_withstringend_with-code-start-code-end

rebased https://github.com/mikel/mail/pull/1231


mikel:
Nice, should be a good speed bump, can you please add a changelog for this and the previous PR as a "performance improvement" change under your name to this PR?

ahorek:
done, but tests are completely broken, probably due to a new psych release, maybe we should lock the previous version for now?

c960657:
@ahorek The Psych issue has now been resolved (see #1522).

PR #1515, opened 2022-12-03
[cherry pick] SMTP: allow disabling starttls_auto since it's now true by default in…

ahorek:
this PR adds changes from https://github.com/mikel/mail/pull/1480 to master branch as requested by https://github.com/mikel/mail/pull/1435#issuecomment-1336121889

### DIFF
--- a/lib/mail/network/delivery_methods/smtp.rb
+++ b/lib/mail/network/delivery_methods/smtp.rb
@@ -135,7 +135,7 @@ module Mail
             when true
               smtp.enable_starttls_auto(ssl_context)
             when false
-              smtp.disable_starttls_auto
+              smtp.disable_starttls
             else
               raise ArgumentError, "Unrecognized :enable_starttls_auto value #{settings[:enable_starttls_auto].inspect}; expected true, false, or nil"
             end
--- a/lib/mail/network/delivery_methods/smtp.rb
+++ b/lib/mail/network/delivery_methods/smtp.rb
@@ -111,17 +111,33 @@ module Mail
 
       def build_smtp_session
         Net::SMTP.new(settings[:address], settings[:port]).tap do |smtp|
-          if settings[:tls] || settings[:ssl]
-            if smtp.respond_to?(:enable_tls)
+          tls = settings[:tls] || settings[:ssl]
+          if !tls.nil?
+            case tls
+            when true
               smtp.enable_tls(ssl_context)
+            when false
+              smtp.disable_tls
+            else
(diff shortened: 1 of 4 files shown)


mikel:
Hey there @ahorek perfect, thank you, can you add the changelog please?  Also, we should check to see if we need to update the README to tell people how to use it.

ahorek:
hey @mikel the changelog entry is already part of 2.8.0 release, see the original PR https://github.com/mikel/mail/pull/1480 , so I don't think it should be duplicated for 2.9.0

there's detailed documentation about available options in Rails https://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-configuration we could add something similar to the README file, but it should be in a separate PR

eval:
@ahorek this code-comment is still relevant: https://github.com/mikel/mail/pull/1435#discussion_r719678459

ahorek:
> @ahorek this code-comment seems still relevant: [#1435 (comment)](https://github.com/mikel/mail/pull/1435#discussion_r719678459)

you're right, thanks

someone could use a setting like this
:enable_starttls      => true,
:enable_starttls_auto => false,

which makes no sense and it'll disable starttls completely, but I guess we need to keep it for compatibility...

ahorek:
@eval fixed

lib/mail/network/delivery_methods/smtp.rb
```diff
@@ -111,17 +111,33 @@ def start_smtp_session(&block)
 
       def build_smtp_session
         Net::SMTP.new(settings[:address], settings[:port]).tap do |smtp|
-          if settings[:tls] || settings[:ssl]
-            if smtp.respond_to?(:enable_tls)
+          tls = settings[:tls] || settings[:ssl]
+          if !tls.nil?
```
eval, line 115: This skips any `enable_starttls*`-setting when `ssl: false`...

eval:
@mikel  I created a new PR that replaces this one to address the inconsistencies @ahorek  and I found.

mikel:
Closing in favour of https://github.com/mikel/mail/pull/1536

akostadinov:
> someone could use a setting like this
> :enable_starttls => true,
> :enable_starttls_auto => false,
> 
> which makes no sense and it'll disable starttls completely, but I guess we need to keep it for compatibility...

It makes sense to make sure that starttls is mandatory. While it may in practice disable it, the setting makes sense and it would be only code that works contrary to reasonable expectation.

PR #1526, opened 2022-12-09
fix undefined method `disable_starttls_auto'

ahorek:
this PR fixes a failure if you pass a setting enable_starttls_auto: false as suggested by https://github.com/mikel/mail/pull/1515#issuecomment-1344455737

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -1,6 +1,7 @@
 == Version 2.8.1 (unreleased)
 
 Bug Fixes:
+* Regression: accept enable_starttls_auto: false @ahorek
 
 == Version 2.8.0 (3-Dec-2022)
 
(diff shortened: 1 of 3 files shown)


mikel:
Potentially handled by https://github.com/mikel/mail/pull/1536 ?

mmustala:
Master branch seems to have this already fixed, only needs a new version released.

ahorek:
I agree with @eval , this one is still valid for 2.8.x. While https://github.com/mikel/mail/pull/1536 also fixes the same thing, it could be a breaking change.

glaszig:
this hit me on production as i just went from 2.7 to 2.8. downgraded & locked to 2.7 for the time being.

ahorek:
> this hit me on production as i just went from 2.7 to 2.8. downgraded & locked to 2.7 for the time being.

as a workaround just remove
**enable_starttls_auto: false**
and keep
**enable_starttls: false**

it should work the same way even without this fix

PR #1532, opened 2022-12-10
Make activesupport gem optional

sebbASF:
It's a nuisance to have to download rails from Github if it is not needed.


mikel:
Nice.

PR #1536, opened 2022-12-11
SMTP: refactor and accept starttls :always and :auto

eval:
This PR is a refactored version of https://github.com/mikel/mail/pull/1515 (which is a master-rebased version of https://github.com/mikel/mail/pull/1435) and adds `:always` and `:auto` as acceptable values for `enable_starttls`.

While the original PR 1515 fixes the disabling of starttls, it also yields unexpected/faulty 'shortcuts' in the if-chain given some inputs, e.g.:
1. `ssl: false` ensures any`enable_starttls`-flag is ignored.
2. `enable_starttls: false` would ignore the `enable_starttls_auto`-flag.

I refactored the logic to come to the right `starttls` value for smtp (i.e. `:always`, `:auto` or `false`). It also adds the option to directly pass `:always` or `:auto` as value for `enable_starttls` (separate commit).

The implementation is based on this decision table:
```
| tls/ssl   | enable_starttls | enable_starttls_auto | result                                     |
|-----------+-----------------+----------------------+--------------------------------------------|
| true      | *               | *                    | {smtp_tls?: true, smtp_starttls: false}^1  |
| - / false | -               | -                    | {smtp_tls?: false, smtp_starttls: :auto}   |
| - / false | - / false       | true                 | {smtp_tls?: false, smtp_starttls: :auto}   |
| - / false | true            | *                    | {smtp_tls?: false, smtp_starttls: :always} |
| - / false | - / false       | false                | {smtp_tls?: false, smtp_starttls: false}   |
| - / false | false           | -                    | {smtp_tls?: false, smtp_starttls: false}   |
| - / false | :always         | *                    | {smtp_tls?: false, smtp_starttls: :always} |
| - / false | :auto           | *                    | {smtp_tls?: false, smtp_starttls: :auto}   |

* = any value
- = not provided (i.e. key absent or nil-value) 

^1: raise when truthy value for any enable_starttls* flag is provided (similar to IMAP).
```

### DIFF
--- a/lib/mail/network/delivery_methods/smtp.rb
+++ b/lib/mail/network/delivery_methods/smtp.rb
@@ -84,7 +84,7 @@ module Mail
       :password             => nil,
       :authentication       => nil,
       :enable_starttls      => nil,
-      :enable_starttls_auto => true,
+      :enable_starttls_auto => nil,
       :openssl_verify_mode  => nil,
       :ssl                  => nil,
       :tls                  => nil,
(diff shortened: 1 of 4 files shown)


mikel:
Hey there @ahorek  and @eval does this also replace https://github.com/mikel/mail/pull/1526 ?

ahorek:
with old rubies (net/smtp), both options were disabled by default, so historically it used to make sense to enable them separately
```
enable_starttls: true
enable_starttls_auto: true
```

but since starttls is now :auto by default due to the upstream change in net/smtp, it's problematic to do the opposite and disable them with existing options

what about introducing a new setting, that accepts the same values?
```
starttls: [false, :auto, :always]
```
if ```enable_starttls``` or ```enable_starttls_auto``` are passed - raise an error with a hint of what the new setting is

this is already a breaking change, so I suggest simplifying it if we have a chance. @mikel @eval What do you think?


> Hey there @ahorek and @eval does this also replace https://github.com/mikel/mail/pull/1526 ?

yes, but the related PR should still be merged for 2.8.x

ron-shinall:
@mikel @eval @ahorek Do you have an ETA for a release that includes this? Thank you!

ahorek:
@ron-shinall I'm not a maintainer.

only @mikel or @jeremy can push a new version, but none of them are active lately

stanhu:
@jeremy Would you mind releasing a new version with this fix? We're seeing a number of users not able to disable STARTTLS without this change.

ahorek:
@stanhu fyi it has already been released in 2.8.1

ahorek:
the main problem was starttls cannot be disabled with any settings, this should already be fixed.

this refactor also fixes a combination of
tls: false + enable_starttls: true (should ignore starttls)
also enable_starttls: true is now enabled by default (in net/smtp) and a simple tls: false doesn't disable it

just pass tls: false + enable_starttls: false that should do what you want. I agree there's still inconsistency, but a workaround is pretty simple.

stanhu:
> a workaround is pretty simple.

Yes, the workaround is simple in principle, but we have potentially thousands of users that have set `settings[:tls]` and `settings[:ssl]` to `false`. We ship mail v2.8.1 now, but having to comment out existing settings that worked is not a good user experience. We will backport this patch for now.

ahorek:
well, this PR is merged for 2.9, but it isn't released. If that's important for you, you should open a new issue and ask maintainers if they're willing to backport it.

lib/mail/network/delivery_methods/smtp.rb
```diff
@@ -105,23 +105,64 @@ def deliver!(mail)
...
+      def smtp_tls?
+        setting_provided?(:tls) && settings[:tls] || setting_provided?(:ssl) && settings[:ssl]
+      end
+
       def start_smtp_session(&block)
         build_smtp_session.start(settings[:domain], settings[:user_name], settings[:password], settings[:authentication], &block)
       end
 
       def build_smtp_session
+        if smtp_tls? && (settings[:enable_starttls] || settings[:enable_starttls_auto])
```
Supermathie, line 148: This change introduces an error condition to a perfectly valid configuration - TLS enabled and opportunistic STARTTLS enabled.

What is the reason for erroring here? I agree with `smtp_tls? && settings[:enable_starttls]` but not `smtp_tls? && settings[:enable_starttls_auto]`.

LucasArruda: I had a broken production (could not send emails due to this new config), apparently

```
ArgumentError
:enable_starttls and :tls are mutually exclusive. Set :tls if you're on an SMTPS connection. Set :enable_starttls if you're on an SMTP connection and using STARTTLS for secure TLS upgrade. (ArgumentError) raise ArgumentError, ":enable_starttls and :tls are mutually exclusive. Set :tls if you're on an SMTPS connection. Set :enable_starttls if you're on an SMTP connection and using STARTTLS for secure TLS upgrade."
```

Fix was to remove both (`enable_starttls_auto: true` and `tls: true`) and set `ssl: true`

Issue #1538, opened 2022-12-13
Sendmail argument change breaks Rails usage of Sendmail


RaymondFallon:
I see from the CHANGELOG for `v2.8.0` that the following change was made intentionally:

> Sendmail and exim delivery :arguments option must be an array of string args

It was even listed as a "breaking change," so this issue is more of a question than a bug report. But I wanted to note that this change breaks [how Rails' ActionMailer implements Sendmail](https://github.com/rails/rails/blob/v7.0.4/actionmailer/lib/action_mailer/delivery_methods.rb#L32):

```ruby
add_delivery_method :sendmail, Mail::Sendmail,
  location:  "/usr/sbin/sendmail",
  arguments: "-i"
```

I don't know whether or not Rails applications represent a large portion of apps that use the `mail` gem, but in case they do, I wanted to point out this potentially high-impact breakage and confirm that it was deemed acceptable before attempting to fix things within Rails. 

Thanks!

--- 

Additional info on breakage:

When attempting to deliver mail via the `:sendmail` method with `rails 7.0.4` and `mail 2.8.0`, I get a runtime error like the following:

```
ArgumentError (:arguments expected to be an Array of individual string args):
  
/gems/mail/lib/mail/network/delivery_methods/sendmail.rb:53:in `initialize'
```

If the Sendmail `argument` signature is to stay as it is, I believe action_mailer will need to change its instantiation of `:sendmail` to:

```ruby
add_delivery_method :sendmail, Mail::Sendmail,
  location:  "/usr/sbin/sendmail",
  arguments: "[-i]"
```

eval:
Seems serious. @mikel how about a fix that doesn't raise here https://github.com/mikel/mail/blob/fbed43b4421d6bb7fed22f76f8b43d1b289b7b3e/lib/mail/network/delivery_methods/sendmail.rb#L53, and instead splits the string (and emits a warning)?

mikel:
@eval yes, that makes sense, we should make that a PR and get that deployed as 2.8.1 right away.  If we get a string, split it and put it back in.

kokogakayui:
Hi.

Just for the sake of feedback : mail 2.8.0 is indeed breaking my rails application in production when trying to deploy on AWS Beanstalk.

I am running ruby 3.0.5 and rails 6.1.7. Reverting back to 2.7.1 fixes it back. 
Did not really find out what the error was, but something related to an uninitialized constant Mail::SendMail.

Hope this will help.

[...]

PR #1544, opened 2022-12-14
Sendmail: allow for some string :arguments

eval:
This PR is a bandaid for callers that initialize Mail::Sendmail with an `arguments` string (https://github.com/mikel/mail/issues/1538).

Rails applications configured to use sendmail and using the default `arguments` from ActionMailer (versions < 7.1), will with mail 2.8.0 experience an ArgumentError. This PR prevents this and shows a deprecation warning:
```
# config in config/environments/$RAILS_ENV.rb
config.action_mailer.delivery_method = :sendmail

# console
Loading development environment (Rails 7.0.4)
irb(main):001:0> UserMailer.welcome_email.deliver_now
DEPRECATION WARNING: Mail::Sendmail :arguments should be an Array of string args, not a string. (called from initialize at /Users/gert/src/mail/lib/mail/network/delivery_methods/sendmail.rb:55)
  Rendering layout layouts/mailer.text.erb
  Rendering user_mailer/welcome_email.text.erb within layouts/mailer
  Rendered user_mailer/welcome_email.text.erb within layouts/mailer (Duration: 0.6ms | Allocations: 280)
  Rendered layout layouts/mailer.text.erb (Duration: 1.1ms | Allocations: 517)
UserMailer#welcome_email: processed outbound mail in 12.8ms
```

A kernel warning is issued when AS is not available - the warning is the same (but without the filename and hopefully descriptive enough).

Callers that have flags other than `-t` and `-i`, will nonetheless experience the ArgumentError (we can't just split the arguments string). But as these applications have explicitly defined these arguments, it should be less cryptic where the errors originate from.

### DIFF
--- a/lib/mail/network/delivery_methods/sendmail.rb
+++ b/lib/mail/network/delivery_methods/sendmail.rb
@@ -49,6 +49,12 @@ module Mail
     end
 
     def initialize(values)
+      if values[:arguments].is_a?(String)
+        values[:arguments] = try_arrayfy_arguments(values[:arguments])
+        if values[:arguments].is_a?(Array)
+          deprecation_warn.call "Mail::Sendmail :arguments should be an Array of string args, not a string."
+        end
+      end
       self.settings = self.class::DEFAULTS.merge(values)
       raise ArgumentError, ":arguments expected to be an Array of individual string args" if settings[:arguments].is_a?(String)
     end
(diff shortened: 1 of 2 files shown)


eval:
Please take a look @mikel @sebbASF

eval:
@sebbASF let's go with that one!

PR #1545, opened 2022-12-14
allow for string :arguments #1544

sebbASF:
Alternative solution without splitting string

### DIFF
--- a/lib/mail/network/delivery_methods/sendmail.rb
+++ b/lib/mail/network/delivery_methods/sendmail.rb
@@ -50,10 +50,11 @@ module Mail
 
     def initialize(values)
       if values[:arguments].is_a?(String)
-        deprecation_warn.call "Mail::Sendmail :arguments should be an Array of string args, not a string."
+        deprecation_warn.call \
+          'Initializing Mail::Sendmail with :arguments of type String is deprecated.' \
+          ' Instead ensure :arguments is an array of strings, e.g. ["-i", "-t"]'
       end
       self.settings = self.class::DEFAULTS.merge(values)
-      # raise ArgumentError, ":arguments expected to be an Array of individual string args" if settings[:arguments].is_a?(String)
     end
 
     def destinations_for(envelope)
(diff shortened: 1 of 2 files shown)


eval:
As we're now passing string-commands again to popen, which go through the shell, this should have the shellquoting reinstated I would say: https://github.com/mikel/mail/commit/7e1196bd29815a0901d7290c82a332c0959b163a

mikel:
Yeah, with the shell quoting, I think we can merge this and do something similar in master too.

lib/mail/network/delivery_methods/sendmail.rb
```diff
@@ -49,8 +49,11 @@ class DeliveryError < StandardError
     end
 
     def initialize(values)
+      if values[:arguments].is_a?(String)
+        deprecation_warn.call "Mail::Sendmail :arguments should be an Array of string args, not a string."
```
eval, line 53: This was my first take on a deprecation message and I think it reads more like an error message.
How about: `Initializing Mail::Sendmail with :arguments of type String is deprecated. Instead ensure :arguments is an array of string args, e.g. ["-i", "-t"].`

eval:
Good stuff @sebbASF 👍🏻

spec/mail/network/delivery_methods/sendmail_spec.rb
```diff
@@ -169,4 +169,17 @@
       mail.deliver!
     end.to raise_error('SMTP To address may not be blank: []')
   end
+
+  it 'allows for legacy callers with :arguments strings containing default flags -i and/or -t' do
+    Mail.defaults do
```
eval, line 174: Adding the following would check for and remove the warning from test-output:
```diff
--- a/spec/mail/network/delivery_methods/sendmail_spec.rb
+++ b/spec/mail/network/delivery_methods/sendmail_spec.rb
@@ -171,6 +171,8 @@ describe Mail::Sendmail do
   end

   it 'allows for legacy callers with :arguments strings containing default flags -i and/or -t' do
+    expect(Kernel).to receive(:warn).with(/:arguments of type String is deprecated/)
+
     Mail.defaults do
       delivery_method :sendmail, :arguments => '-i -t'
     end
```

eval:
Squashed and merged 🚀

PR #1400, opened 2020-06-30
Remove unused variables warnings

GQuirino:
This PR fixes #1384 removing warnings from unused variables.

PR #1551, opened 2022-12-23
Fix all 'assigned but unused variable' warnings

skipkayhil:
- The variables in content_location_parser appear to have been copied from content_disposition_parser but are unused
- Ragel seems to always generate a useless testEof variable. The `if false` trick is also used in https://github.com/whitequark/parser/commit/21581a2 to suppress the warning. As mentioned in the parser commit, `if false` will not actually generate instructions and so it has no runtime cost.

The first commit is simply running `rake ragel` again to regenerate the parsers

Fixes #1400 (I've added @GQuirino as a Co-Author)

### DIFF
--- a/lib/mail/parsers/address_lists_parser.rb
+++ b/lib/mail/parsers/address_lists_parser.rb
@@ -1,4 +1,3 @@
-
 # frozen_string_literal: true
 require "mail/utilities"
 require "mail/parser_tools"
@@ -32011,10 +32010,10 @@ begin
               _trans = if (_slen > 0 &&
                            _trans_keys[_keys] <= _wide &&
                            _wide <= _trans_keys[_keys + 1])
-                         _indicies[_inds + _wide - _trans_keys[_keys]]
-                       else
-                         _indicies[_inds + _slen]
-                       end
+                  _indicies[_inds + _wide - _trans_keys[_keys]]
+                else
+                  _indicies[_inds + _slen]
+                end
               cs = _trans_targs[_trans]
               if _trans_actions[_trans] != 0
                 case _trans_actions[_trans]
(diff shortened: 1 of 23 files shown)


skipkayhil:
I tried to wrap my head around the `statement not reached` warnings but haven't cracked it.

An example is this:
```ruby
when 17
  # line 24 "lib/mail/parsers/rfc5322_lexical_tokens.rl"
  begin
    begin
      stack[top] = cs
      top += 1
      cs = 20
      _goto_level = _again
      next
    end
  end
  # line 20 "lib/mail/parsers/content_location_parser.rl"
  begin
    content_location.location = chars(data, token_string_s, p - 1)
  end
```

The comments indicate that the first block is generated by `comment_begin`, and the second block by `token_string_e`.

Mostly just dumping some thoughts, I think this PR should be merged whether or not the other warnings are addressed.

skipkayhil:
> It might be better to set things up so warnings are not generated for these files.

Since the `parser` gem uses the same trick to fix the `testEof` warning, I would be surprised if there was a better way. The rest of the warnings are actually unused variables, so removing them fixes the warnings.

I opened #1557 as another potential method to silence warnings, but someone pointed out to me that changing a global variable in that way is not thread safe. 

Whether or not we find a solution to unreachable statement warnings, I think this change is worth merging as it is an improvement on the current situation.

eval:
Thanks!

mperham:
Any chance we can get a 2.8.2 patch release? It's been almost two years since 2.8.1 was released and QOL fixes like this are lingering, unreleased.

eval:
@mperham keep an eye on https://github.com/mikel/mail/issues/1604

Issue #1489, opened 2022-05-17
Incorrect file permissions in 2.8.0


CvX:
Some files in 2.8.0.rc1 release are not world-readable: (notice `elements.rb`, `fields.rb`, `indifferent_hash.rb`)

```
-rw-r--r--@  1 cvx  staff   3481 Apr 22 02:41 attachments_list.rb
-rw-r--r--@  1 cvx  staff   9545 Apr 22 02:41 body.rb
-rw-r--r--@  1 cvx  staff   1781 Apr 22 02:41 configuration.rb
-rw-r--r--@  1 cvx  staff   2472 Apr 22 02:41 constants.rb
drwxr-xr-x  14 cvx  staff    448 May 17 16:51 elements
-rw-r-----@  1 cvx  staff    960 Apr 22 02:41 elements.rb
drwxr-xr-x  10 cvx  staff    320 May 17 16:51 encodings
-rw-r--r--@  1 cvx  staff   9792 Apr 22 02:41 encodings.rb
-rw-r--r--@  1 cvx  staff    573 Apr 22 02:41 envelope.rb
-rw-r--r--@  1 cvx  staff   8885 Apr 22 02:41 field.rb
-rw-r--r--@  1 cvx  staff   2077 Apr 22 02:41 field_list.rb
drwxr-xr-x  41 cvx  staff   1312 May 17 16:51 fields
-rw-r-----@  1 cvx  staff   2244 Apr 22 02:41 fields.rb
-rw-r--r--@  1 cvx  staff   7528 Apr 22 02:41 header.rb
-rw-r-----@  1 cvx  staff   3874 Apr 22 02:41 indifferent_hash.rb
-rw-r--r--@  1 cvx  staff   8346 Apr 22 02:41 mail.rb
drwxr-xr-x   4 cvx  staff    128 May 17 16:51 matchers
-rw-r--r--@  1 cvx  staff  67251 Apr 22 02:41 message.rb
drwxr-xr-x   5 cvx  staff    160 May 17 16:51 multibyte
-rw-r--r--@  1 cvx  staff   3563 Apr 22 02:41 multibyte.rb
drwxr-xr-x   4 cvx  staff    128 May 17 16:51 network
-rw-r--r--@  1 cvx  staff    836 Apr 22 02:41 network.rb
-rw-r--r--@  1 cvx  staff    446 Apr 22 02:41 parser_tools.rb
drwxr-xr-x  34 cvx  staff   1088 May 17 16:51 parsers
[...]

PR #1554, opened 2023-01-12
Rubygems release on tag push

benlangfeld:
It is known that there is a defect in the release process when triggered in a certain context, evident from https://github.com/mikel/mail/issues/1489. This results in a large degree of frustration, as well as the inability of anyone other than the individual who triggered the release to debug the source of the problem.

Here we automate the release process as a Github Action which is triggered by a push of a tag which begins with `v`, such as `v2.8.1`. In order for this to work correctly, it is necessary to set the secret `rubygems_api_key` on the Github repo, which is something only @mikel can do, but once he does that a release can be cut by anyone with the ability to push tags to this repo, and the release process is executed in a clean context like that in which tests are run, making it available to debugging efforts by the broader community.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,6 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches: [ master ]
+    tags: [v*]
   pull_request:
     branches: [ master ]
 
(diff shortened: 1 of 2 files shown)


benlangfeld:
@sebbASF This works well for my organisation on several projects, and we've never had a problem with mistakenly creating a tag; we use the Github Releases UI to create tags after merging a PR which bumps the gemspec version. I get that in theory a mistaken push could happen, but in my experience it's so rare (I've never seen it in hundreds of releases across dozens of projects) that it's not worth worrying about.

> Is there a way to generate a staging release that can be used for testing and then promoted to a formal release?

That does not address the concern which arises out of #1489 , I'm not sure what the benefit would be.

benlangfeld:
> > That does not address the concern which arises out of #1489 , I'm not sure what the benefit would be.
> 
> I was thinking that a staging release would be used in a test suite which is set up to run tests as a user id other than the installer. This would have detected the permission error. If the same artifact can be promoted to a formal release, then it would already have been tested. Obviously if the release has to be regenerated, that assurance would be lost.

Indeed, but that's far more specific than I was attempting to resolve here. I'm not trying to build any kind of regression testing for the permissions issue, because it's very specific to @mikel 's machine/routine, I'm just trying to generally decouple the release process from his personal situation and put it somewhere repeatable.

As for detecting the problem in #1489 ahead of time, it was detected several RC releases in advance and was not considered in the final release, so I don't think any more automation of the detection of the issue would have helped.

benlangfeld:
@eval Interested in moving forward with this or should I close it?

PR #1556, opened 2023-01-12
Run CI against release branches and tags

jeremy:
And drop defunct workflow used to support #1511


jeremy:
Screenshot 2023-01-12 at 12 19 50 PM
😂

PR #1559, opened 2023-01-23
Revert PR #1495 because it is a dupe of #1470

johnnyshields:
Partial revert of PR #1495 ("Preserve message-level charset when adding parts") because it was fixed by PR #1470 ("Adding a part should not reset the mail's charset to nil") merged around the same time. The fix in #1470 is more elegant.

The specs added in #1495 verify that it works with the #1470 fix.

### DIFF
--- a/lib/mail/message.rb
+++ b/lib/mail/message.rb
@@ -2073,11 +2073,7 @@ module Mail
 
     def add_boundary
       unless body.boundary && boundary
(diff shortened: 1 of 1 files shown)


johnnyshields:
@mikel please merge?

johnnyshields:
Thank you!

PR #1577, opened 2023-04-18
Backport 1536

eval:
This backports #1536 to the `2-8-stable`-branch.

## Why

Though `mail` v2.8 enabled turning TLS/STARTTLS options off (this surfaced as net-smtp's default value for `starttls` went from `false` to `:auto`), it also introduced some new problems:
- having `ssl` `false` would ignore any starttls-setting
- `enable_starttls: false` ignores the `enable_starttls_auto`-flag. 

These issues were fixed on mainline but I think it's worth to have these in a new 2.8 release as well.  

## Impact

Possible outcomes:
- having either `ssl` or `tls` truthy ánd having truthy values for any of `enable_starttls*`-flags set will now raise an ArgumentError.
  While `mail` is stricter in this regard, net-smtp would already consider [this a non-acceptable config](https://github.com/ruby/net-smtp/blob/v0.1.0/lib/net/smtp.rb#L344).
- another scenario: a config with `enable_starttls false` and `enable_starttls_auto true` resulted in `starttls :auto` <v2.8, in v2.8 this resulted in `starttls false` and would now again result in`:auto`.

Did I miss anything?

### DIFF
--- a/lib/mail/network/delivery_methods/smtp.rb
+++ b/lib/mail/network/delivery_methods/smtp.rb
@@ -22,7 +22,7 @@ module Mail
   #                              :user_name            => '<username>',
   #                              :password             => '<password>',
   #                              :authentication       => 'plain',
-  #                              :enable_starttls_auto => true  }
+  #                              :enable_starttls      => :auto  }
   #   end
   #
   # === Sending via GMail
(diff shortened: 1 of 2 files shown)


eval:
@ahorek @sebbASF review appreciated.

eval:
Merging as this was reviewed/discussed in the master-PR.

PR #1587, opened 2023-09-20
Stop using insecure deprecated Net::IMAP.new args

### LINKED ISSUE #1586: Net::IMAP connection uses insecure deprecated options
The parameters that are used to create the Net::IMAP connection in Mail::IMAP are undocumented:
https://github.com/mikel/mail/blob/b6b6cb737d47a85ddc720fda0e6b991e99224848/lib/mail/network/retriever_methods/imap.rb#L167

I believe the documentation was removed in 2007 (ruby/net-imap@e631911e).  Pending the following PR: ruby/net-imap#175, the next release of net-imap will print a deprecation warning when those parameters are used.  About a year later, they'll be removed entirely.

A bigger issue is that the fifth and final argument is false: that is the `verify` parameter and by setting it to false, the SSLContext will have `verify_mode: OpenSSL::SSL::VERIFY_NONE`.

### Proposal

When `Hash.try_convert(enable_ssl)` returns a Hash, that will be used with
```ruby
  Net::IMAP.new(settings[:address],
                port: settings[:port],
                ssl: Hash.try_convert(settings[:enable_ssl]))
```

When `Hash.try_convert(enable_starttls)` returns a Hash, that will be used with
```ruby
  imap.starttls Hash.try_convert(settings[:enable_starttls])
```

In this way, _any_ SSLContext attribute can be set.

When `enable_ssl` or `enable_starttls` are truthy but not hash-like, use `{}`.  This way, the default is secure.

I can write a PR for this, if you agree with my proposal.

nevans:
This fixes the insecure `verify = false` argument that was previously unconfigurable.  Now _any_ SSLContext params can be set on the IMAP connection.

The original parameters have been undocumented since ~2007, will print deprecation warnings in the next release and will be removed in a year.

Fixes #1586.

### DIFF
--- a/lib/mail/network/retriever_methods/imap.rb
+++ b/lib/mail/network/retriever_methods/imap.rb
@@ -167,9 +167,13 @@ module Mail
           raise ArgumentError, ":enable_starttls and :enable_ssl are mutually exclusive. Set :enable_ssl if you're on an IMAPS connection. Set :enable_starttls if you're on an IMAP connection and using STARTTLS for secure TLS upgrade."
         end
 
-        imap = Net::IMAP.new(settings[:address], settings[:port], settings[:enable_ssl], nil, false)
+        ssl      = settings[:enable_ssl]
+        starttls = settings[:enable_starttls]
+        ssl      &&= Hash.try_convert(ssl)      || {}
+        starttls &&= Hash.try_convert(starttls) || {}
 
-        imap.starttls if settings[:enable_starttls]
+        imap = Net::IMAP.new(settings[:address], port: settings[:port], ssl: ssl)
+        imap.starttls(starttls) if starttls
 
         if settings[:authentication].nil?
           imap.login(settings[:user_name], settings[:password])
--- a/spec/mail/network/retriever_methods/imap_spec.rb
+++ b/spec/mail/network/retriever_methods/imap_spec.rb
@@ -269,6 +269,27 @@ RSpec.describe "IMAP Retriever" do
     end
   end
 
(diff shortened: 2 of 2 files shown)


nevans:
How should this change be documented?

sebbASF (approved):
Looks good to me

eval:
I think accepting ssl_context_params as value for `enable_(ssl|starttls)` is somewhat opaque. It's also different from how smtp is configured.

As of branch `2-8-stable` this is the situation:
```
| goal            | smtp                                                                  | imap                   |
|-----------------+-----------------------------------------------------------------------+------------------------|
| enable tls/ssl  | options tls (bool), ssl (bool)                                        | enable_ssl (bool)      |
| enable starttls | enable_starttls (bool)                                                | enable_starttls (bool) |
| verify certs    | openssl_verify_mode (string or integer) (serving tls and starttls)    |                        |
| certs           | (undocumented) options ca_path and ca_file (serving tls and starttls) |                        |
```

How about keeping the `enable_*` options as is and adding the options `openssl_verify_mode`, `ca_path` and `ca_file` just like smtp?

> Fixing security bugs this severe should not cause a major version bump for the simple reason that a major version bump will prevent users from receiving the security fix.
https://github.com/mikel/mail/issues/1586#issuecomment-1754237729

As for the 'reach' of this secure-by-default-setting: I think the `2-8-stable`-branch is the one to target alongside the master-branch. The two branches have diverted in such a way (i.e. a lot of 2-8-stable is not on master) that I don't see a release being cut from master any time soon.

PR #1605, opened 2024-02-29
Update documentation regarding errors array

mikehale:
The errors list is built in field.rb like this:

```
      field.errors << [name, value, e]
```

That means there are 3 elements, but the last one would be at index 2, not 3.


eval (approved):

eval:
Thanks!

PR #1613, opened 2024-06-07
Document SMTP TLS/STARTTLS settings (cherry-picked from 2.8 stable branch)

nevans:
This PR simply cherry-picks 3f05820b from #1577 onto the main branch.  It looks to me like the 2.8 docs were updated after #1536, and missing them in `master` was probably unintentional.


eval (approved):
Good catch 👍🏻

PR #1623, opened 2024-10-28
[rebased] Use ragel-bitmap for less memory

ahorek:
this is a rebased version of https://github.com/mikel/mail/pull/1343

it significantly reduces resident memory by -39MB
but there are downsides:

performance
     -9% mri 3.4.0 (dev)
     -1% mri 3.4.0+yjit
     -10% jruby
       0% jruby+indy
      -10% truffleruby
and +6% runtime allocations

however, it's a micro-benchmark and differences are marginal, so it's unlikely to be a bottleneck in a real application.


jrafanie:
@ahorek can you share your benchmark script?  I'd like to test in my project but am not sure if the parsers are eager required via `require 'mail'` or if I need to do something different to demonstrate the benefits and tradeoffs of this change.

Thanks!

ahorek:
nothing special
```
file = File.binread("email.eml")
Mail.new(file).to_s
```

https://github.com/evanphx/benchmark-ips
https://github.com/zombocom/derailed_benchmarks

Issue #1342, opened 2019-05-29
address_lists_parser.rb requires a lot of memory (~30 mb)


schneems:
Loading the `address_lists_parser.rb` is the cause for about 1/3 of the startup memory for CodeTriage.

```
TOP: 105.7266 MiB
  mail/parsers: 31.9609 MiB
    mail/parsers/address_lists_parser: 29.7266 MiB
```

> Using the `derailed_benchmarks` gem on codetriage/codetriage.

It looks like it is generated by a parser generator Is there some way we can reduce that memory overhead?

ahorek:
There's an option to recompile ragel for lower memory consumption, but it does hurt performance, see https://github.com/mikel/mail/issues/1215

@jeremy any plans to support ragel 7?

schneems:
The Ragel Bitmap PR looks pretty promising. Doesn’t look like it affects performance.

I tested with derailed and instead of 30 mb I’m seeing about 5mb. Which is a pretty significant improvement.

jeremy:
Remember #812 → #815 → #912? 😅

Lazy-loading the parsers is desirable, considering that most usage is building and delivering messages, not parsing them. But the lazy-loading has to work consistently.

ahorek:
Well, it depends. For our helpdesk we use the mail gem for parsing incoming messages a lot.
Ragel never used to generate a performant code in ruby, but the question is if 30MB per instance really matters these days...

schneems:
😅 Sorry for the duplicate issue spam. I was working with someone on derailed benchmarks and mis-remembered that somehow I had gotten away without the memory increase for https://www.codetriage.com. 

While it doesn't look like there is an obvious replacement (that satisfies all memory and speed requirements) it did generate some good discussion and eyeballs in #1343 whether any changes end up getting merged or not, there's some new techniques in there that i've not seen before.

> the question is if 30MB per instance really matters these days...

It matters more for memory constrained implementations, FaaS and PaaS. Usually, there's a line somewhere and if you're over even a few mb, you'll still start to swap and incur enormous perf penalties. 

I certainly understand that for some absolute performance is important. I also understand that one of the goals of the project is to not have a c-extension dependency. 

Thanks for your time and maintenance!

gatopanx:
any workarounds for Rails , seeing the same:

derailed output:
```
TOP: 175.1016 MiB
  rails/all: 58.0781 MiB
    action_mailbox/engine: 31.4609 MiB
[...]

PR #1624, opened 2024-10-29
use ragel-bitmap for less memory

eval:
port of #1623


eval:
@ahorek it seems [received_parser.rb](https://github.com/mikel/mail/blob/64b5d722b3926ccf0f1e73b1131e567ba99422fc/lib/mail/parsers/received_parser.rb) was unchanged in the ragel-bitmap PR. This makes sense as the timestamps of the rl-files hadn't changed.
Manually deleting `lib/mail/parsers/*.rb` and re-running `rake ragel:generate` revealed it.

Could you approve this PR so it can be merged? I'll add the extra commit to 2-8-stable.
To clean up these files I created #1626 - review welcome as well 🙏🏻

ahorek:
Nicely spotted! Thanks for catching that.

ahorek (approved):
👍

eval:
The approval needs to be from someone with commit rights, it seems... mind to help out with this ported PR @sebbASF ?

eval:
@sebbASF how would you suggest to move fw?

ahorek:
Compared to other dependencies, this gem has a disproportionately large memory footprint for a simple email parser. This is primarily because Ragel doesn’t generate efficient code for Ruby, and Ruby doesn’t handle large numeric arrays optimally.

https://github.com/mikel/mail/issues/1342 cc @schneems @kddnewton

I don’t believe this complicates the build process. Modifying the Ragel sources requires recompilation regardless, and there’s already a rufo formatter in place. Additionally, the parsers were last updated in 2017, so this change shouldn’t increase maintenance costs.

kddnewton:
I think this would be a worthwhile change, but ultimately I wouldn't be the one maintaining it so I don't think I can weigh in very heavily. I believe the memory savings are good, but if it's too complicated to understand maybe it's not worth it.

radar:
I think the addition of benchmarks in this PR would be a worthwhile addition, either in the original PR description or as a comment here. Show us your data :)

fz-ryanbigg:
This may have already been merged in https://github.com/mikel/mail/pull/1623.

radar:
This may have already been merged in #1623.

ahorek:
It was only merged into the 2.8 branch except for https://github.com/mikel/mail/pull/1630
If you intend to merge 2.8 into master manually, this PR may not be needed.

Issue #1627, opened 2024-12-05
Parsing mail with attachment that has no filename will not find attachment


lazyatom:
I've been parsing emails from from Outlook (I think), and I've seen that the attachment parts have no filename information:

```
# ... earlier parts of email

--_001_ABC123MAILPARTDIVIDERRANDOMNUMBERS_
Content-Type: message/rfc822
Content-Disposition: attachment;
	creation-date="Tue, 03 Dec 2024 12:01:37 GMT";
	modification-date="Tue, 03 Dec 2024 12:01:37 GMT"

From: Alice Smith <alice@example.com>
To: "Bob" <bob@example.com>
Subject: Some email file I'm attaching

# etc ...
```

The Mail gem fails to recognise this as an attachment, despite the content disposition, because there's nothing it recognises as the filename, and that's fundamental to how Mail determines if an attachment exists.

According to the RCF, the [filename is not mandatory](https://datatracker.ietf.org/doc/html/rfc2183#section-2), and I believe the intention that mail implementations _can_ use the value, if present, but also aren't obliged to.

In my particular case, I'd like to be able to get all parts that have `Content-Disposition: attachment` as attachments, and I am not sure it's technically correct that our gem only reports parts that have filenames. I would be happy to supply a default filename; perhaps `message.attachments` would expose the nameless attachment's mime type to me and let me use something like [Marcel](https://github.com/rails/marcel) to determine an appropriate file extension.

I'd be more than happy to work on a PR; I have a patch in my app for the particular case I'm hitting, but if we can figure out a general approach that works for all unnamed attachment types, I'd be more than happy to put the work in to address it.

ahorek:
could you check if https://github.com/mikel/mail/pull/1389 fixes the issue? the change hasn't been released yet.

lazyatom:
That works for my test case, certainly!

PR #1628, opened 2024-12-17
include attachments without a filename

ahorek:
fixes https://github.com/mikel/mail/issues/1627


abuisman:
@mikel can we rerun the checks that failed here? Maybe we can help out fixing them or is it not something you'd want to merge?

PR #1637, opened 2025-03-27
Fix JRuby builds / minimum ruby version

eval:
Recent JRuby builds were failing with errors "Error: Unavailable version 9.2.21.0 for jruby on ubuntu-24.04".

This PR ensures JRuby uses ubuntu-22.04.

It also prunes builds for CRuby 2.5 and 2.6.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   ruby:
     name: ${{ matrix.ruby }} (timeout ${{ matrix.timeout }})
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


eval:
As Ruby 2.7 is now [2 years EOL](https://www.ruby-lang.org/en/downloads/branches/), maybe a good moment to prune that version as well, WDYT @radar?

fz-ryanbigg:
Yep, trim it. While you're there, also trim 3.0 (EOL'D almost a year ago). Keep 3.1, despite it EOL'ing in 4 days time.

eval:
@fz-ryanbigg 3.1 it is 👍🏻

eval:
@fz-ryanbigg good to go?

PR #1640, opened 2025-06-12
Remove unusued `testEof` variables

viralpraxis:
There's a bunch of weird `testEof` variable declarations with the same pattern:

```ruby
testEof = false

...

if false
  testEof
end
```

these are leading to runtime warnings, so I guess we can simply remove them.

### DIFF
--- a/lib/mail/parsers/address_lists_parser.rb
+++ b/lib/mail/parsers/address_lists_parser.rb
@@ -31983,7 +31983,6 @@ begin
         end
 
         begin
-          testEof = false
           _slen, _trans, _keys, _inds, _acts, _nacts = nil
           _goto_level = 0
           _resume = 10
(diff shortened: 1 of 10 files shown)


skipkayhil:
These files are generated, so should probably not be manually modified.

The `if false; testEof; end` was added to _fix_ unused variable warnings in https://github.com/mikel/mail/pull/1551. What warnings are you seeing now?

viralpraxis:
> These files are generated, so should probably not be manually modified.
> 
> The `if false; testEof; end` was added to _fix_ unused variable warnings in #1551. What warnings are you seeing now?

You're absolutely right, it's just that I didn't bother to check if warnings are fixed on the latest master commit. Seems like the problem here is that there's no release yet, so I'm closing this. Thanks you for responding.

